PR #23, opened 2022-11-04
feat: rewrite Button component like Typography

Rexogamer:
(and made a few misc fixes too)

### DIFF
--- a/components/ui/components/design/atoms/inputs/Button.tsx
+++ b/components/ui/components/design/atoms/inputs/Button.tsx
@@ -1,6 +1,3 @@
-// ! TODO: this is a direct port of the Button
-// ! this will need to be polished up and rewritten for solid-styled-components
-
 import type { JSX } from "solid-js/jsx-runtime";
 import { styled } from "solid-styled-components";
 
@@ -50,8 +47,7 @@ export interface Props {
 /**
  * Common button styles 
 */
-function buttonBase (compact: boolean | "icon") {
-  return `
+const ButtonBase = styled("button")<Props>`
   z-index: 1;
 
   display: flex;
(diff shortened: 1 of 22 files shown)


insertish:
recheck

components/ui/components/design/atoms/inputs/Button.tsx
```diff
@@ -24,9 +25,33 @@ export interface Props {
...
+    padding: 2px 16px;
+    font-size: 0.8125rem;
+  `
+}
+
+
+/**
+ * Common button styles 
+*/
+function buttonBase (compact: boolean | "icon") {
```
insertish: You can convert `buttonBase` into `ButtonBase` with is a styled component:
```js
const ButtonBase = styled.button``;
```

Then inherit from it in the other places it is needed:
```js
const PrimaryButton = styled(ButtonBase)``;
```

insertish (changes requested):
Almost there, we can save a bit on classes here though.

components/ui/components/navigation/channels/HomeSidebar.tsx
```diff
@@ -98,7 +99,7 @@ function Entry({
 export const HomeSidebar = ({ conversations, channelId }: Props) => {
   return (
     <SidebarBase>
-      <p>
+      <p style={"margin-left: 0.5em;"}>
```
insertish: also this shouldn't be here or in this PR

components/ui/components/messaging/message/Message.tsx
```diff
@@ -30,7 +31,7 @@ export function Message({
         </Show>
       }
     >
-      {message.content}
+      {content}
```
insertish: likewise

insertish (changes requested):
Added two things

insertish (approved):

PR #188, opened 2023-04-23
Feat: including "replying to" in copy text (resolve #172)

### LINKED ISSUE #172: feat: include "replying to" in copy text

kimlimjustin:

---
Tricking copying "replying to" text into the keyboard by adding an invisible element, as suggested at #172. Hope this changes look good :)

### DIFF
--- a/components/ui/components/messaging/message/MessageReplyCopyHelper.tsx
+++ b/components/ui/components/messaging/message/MessageReplyCopyHelper.tsx
@@ -1,0 +1,25 @@
+import { styled } from "solid-styled-components";
+
+import { Message as MessageInterface } from "revolt.js";
+
+export const ReplyHelperComponent = styled("span")`
+  opacity: 0;
+  pointer-events: auto;
+  z-index: -1;
+  height: 0;
+  margin: 0 -3px;
+  display: block;
+`;
+
+/**
+ * Helper component to copy the reply by using invisible text
+ * @param props
+ */
+export function MessageReplyCopyHelper(props: { message: MessageInterface }) {
+  // make the element invisible but selectable
+  return (
+    <ReplyHelperComponent>
+      Replying to {props.message.username}: {props.message.content}
+    </ReplyHelperComponent>
+  );
+}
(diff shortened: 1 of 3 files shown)


kimlimjustin:
I have read the CLA Document and I hereby sign the CLA

kimlimjustin:
recheck

insertish:
Would it not be possible to include hidden text directly in the `MessageReply` component rather than refactoring the Message component? (and duplicating some code)

kimlimjustin:
> Would it not be possible to include hidden text directly in the `MessageReply` component rather than refactoring the Message component? (and duplicating some code)

Sorry for the late response... I've pushed a commit, could you please recheck it?

components/ui/components/messaging/message/MessageReply.tsx
```diff
@@ -87,43 +89,48 @@ export function MessageReply(props: Props) {
...
-            >
-              <Typography variant="username">
-                {props.mention && "@"}
-                {props.message!.username}
-              </Typography>
-            </ColouredText>
-          </NonBreakingText>
-          <Show when={props.message!.attachments}>
+    <>
+      <MessageReplyCopyHelper message={props.message!} />
```
insertish: If `props.message` is not yet defined then this would crash out.

components/ui/components/messaging/message/MessageReply.tsx
```diff
@@ -87,43 +89,48 @@ export function MessageReply(props: Props) {
...
+                <Attachments>
+                  <BiSolidFile size={16} />
+                  {props.message!.attachments!.length > 1
+                    ? t("app.main.channel.misc.sent_multiple_files")
+                    : t("app.main.channel.misc.sent_file")}
+                </Attachments>
+              </NonBreakingText>
+            </Show>
+            <Show when={props.message!.content}>
+              <OverflowingText>
```
insertish: Invisible text could be placed above this with user-select enabled.
(the reply span styled component would also be defined in this file)

insertish (changes requested):
Sorry currently doing exams but, there is still quite a large diff on this PR.

There is at least one issue reading this that could cause a crash but beyond that I would probably refactor this further into just one file, in theory you should be able to just define the "ReplyHelperComponent" styled span component in MessageReply itself and then use it as appropriate.

insertish:
PR went stale, feedback not responded to, closing out for now

PR #250, opened 2023-07-30
feat(modal): Port ServerIdentity (closes #112)

### LINKED ISSUE #112: feat(modal): port ServerIdentity

aisuleep:

### DIFF
--- a/components/modal/modals/ServerIdentity.tsx
+++ b/components/modal/modals/ServerIdentity.tsx
@@ -1,0 +1,345 @@
+import { BiRegularX, BiSolidEditAlt, BiSolidSave } from "solid-icons/bi";
+import { Show, createSignal } from "solid-js";
+
+import { ServerMember } from "revolt.js";
+
+import { clientController } from "@revolt/client";
+import { useTranslation } from "@revolt/i18n";
+import { userInformation } from "@revolt/markdown/users";
+import {
+  Avatar,
+  Button,
+  CategoryButton,
+  Column,
+  Input,
+  NonBreakingText,
+  OverflowingText,
+  Preloader,
+  Row,
+  Typography,
+  Username,
+  styled,
+  useTheme,
+} from "@revolt/ui";
+import { generateTypographyCSS } from "@revolt/ui/components/design/atoms/display/Typography";
+
+import { PropGenerator } from "../types";
+
+const [avatarId, setAvatarId] = createSignal(null);
+const [uploading, setUploading] = createSignal<boolean>();
+
+async function uploadToAutumn(file) {
+  setUploading(true);
+  try {
+    const url = "https://autumn.revolt.chat/avatars";
+    const formData = new FormData();
+    formData.append("file", file);
+
+    const response = await fetch(url, {
+      method: "POST",
+      headers: {
+        "x-session-token": `${clientController.getCurrentClient()?.sessionId}`,
+      },
+      body: formData,
+    });
+
+    const json = await response.json();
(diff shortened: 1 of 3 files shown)


aisuleep:
Not sure if still needed, but the issue was still labeled as "help-wanted" so I spruced up and finished the draft PR from before. :+1:

aisuleep:
I have read the CLA Document and I hereby sign the CLA

components/modal/modals/ServerIdentity.tsx
```diff
@@ -0,0 +1,345 @@
...
+  useTheme,
+} from "@revolt/ui";
+import { generateTypographyCSS } from "@revolt/ui/components/design/atoms/display/Typography";
+
+import { PropGenerator } from "../types";
+
+const [avatarId, setAvatarId] = createSignal(null);
+const [uploading, setUploading] = createSignal<boolean>();
+
+async function uploadToAutumn(file) {
```
amycatgirl, line 31: You should use `File | Blob` instead of not providing a type (aka. implicit `any`) since the `FileList` type is just an array of `File`s (See MDN's article on FileList: https://developer.mozilla.org/en-US/docs/Web/API/FileList)
```suggestion
async function uploadToAutumn(file) {
```

components/modal/modals/ServerIdentity.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const user = () => userInformation(props.member.user, props.member);
+
+  let fileInputRef: HTMLInputElement | null = null;
+  const openFilePicker = () => {
+    fileInputRef?.click();
+  };
+
+  const handleFileSelect = async (event: Event) => {
+    const file = event.target?.files[0];
+    const json = await uploadToAutumn(file);
```
amycatgirl, line 214: The `target` property in `Event` isn't guaranteed to be `HTMLInputElement`, so typescript doesn't know whether the `files` property actually exists or not and throws a error.

```suggestion
  const handleFileSelect = async (event: Event & {target: HTMLInputElement}) => {
```

components/modal/modals/ServerIdentity.tsx
```diff
@@ -0,0 +1,345 @@
...
+  setUploading(false);
+
+  let fileInputRef: HTMLInputElement | null = null;
+  const openFilePicker = () => {
+    fileInputRef?.click();
+  };
+
+  const handleFileSelect = async (event: Event) => {
+    const file = event.target?.files[0];
+    const json = await uploadToAutumn(file);
```
amycatgirl, line 74:
```suggestion
  const handleFileSelect = async (event: Event & {target: HTMLInputElement}) => {
```

components/modal/modals/ServerIdentity.tsx
amycatgirl, line 214: Typescript says that `files` can be null, so it's better to use the non-null assertion operator to tell it to shut up.

```suggestion
    const file = event.target?.files![0];
```

components/modal/modals/ServerIdentity.tsx
amycatgirl, line 74:
```suggestion
    const file = event.target.files![0];
```

amycatgirl (changes requested):
I've left some suggestions regarding types in some functions in `ServerIdentity.tsx`

Other than that, LGTM

insertish:
PR went stale, feedback not actioned / responded to, closing out for now
NB. conflicts with draft ServerIdentity coming in from M3 branch

PR #263, opened 2024-03-16
refactor: the flattening

insertish:
get rid of monorepo (for components), it was a stupid idea


insertish:
ok i have to do this again because upgrading was stpid and i didnt move the submodule correctly

PR #285, opened 2024-06-04
feat: include replies in copied text

### LINKED ISSUE #172: feat: include "replying to" in copy text

JackDotJS:

this PR will resolve #172 

currently a draft while i sort out some edge cases and locale stuff, but the basic functionality is there already.


sharkaccino:
ran into some formatting issues with solid components so i figured i'd put all the logic into its own function instead. i think it ends up being much more readable as a result anyway, but i can revert it if solid components would be preferred.

aside from that, im also wondering if i should just expand to PR to fix copy text in general since there seems to be a number of issues outside of just missing the "replying to" text.

sharkaccino:
gonna be closing this for now. this is super old, fails some checks, and doesn't comply with new DCO thing.

the changes as-is are pretty minimal though, so i can probably revisit it soon.

PR #301, opened 2024-07-17
fix #299

Rushmore75:
## Please make sure to check the following tasks before opening and submitting a PR

- [x] I understand and have followed the [contribution guide](https://github.com/revoltchat/.github/blob/master/.github/CONTRIBUTING.md) (probably, signing old commits is too hard)
- [X] I have tested my changes locally and they are working as intended

remove references to `wordmark_wide_500px.svg` and replace them with `wide.svg`

PR #302, opened 2024-07-17
Env Variable Support

Rushmore75:

Add support for more environment variables in a similar fashion to how to old system works.
https://github.com/revoltchat/self-hosted/pull/90 Is related, it shows the .env config already.

(The fact that my branch is called docker is because it was going to be to dockerize this, then I realized that the self-host stack already has Caddy in it so that can be used to host the dist folder)


insertish:
I'm not sure what's going on with this and #301, but there's a lot of unrelated changes strewn about.
You may want to re-create your PRs to only modify what you're intending to modify.

Rushmore75:
It's pretty much just changing the environment variables to more specific ones, instead of using one env variable for the whole project for url

Then updating the Readme so it reflects the changes + removing parts of the Readme where it references repos that no longer exist

Also fixing a file name for an svg that no longer exists

Also fixing the package.json to have a command to build for pro

insertish:
Closing out as a lot of these changes are unneeded, break process, or break production deployment (which some of these commands are only intended for).

PR #333, opened 2024-10-26
fix(messaging): fix several linebreak issues

JackDotJS:
This PR fixes several minor oddities regarding the behavior of linebreaks, including:
- empty rows of cells at the end of tables when they're followed by further content afterwards
- weird unicode characters appearing between each linebreak in multiline codeblocks
- general inconsistencies with the number of linebreaks between certain markdown elements

<strike>

This PR *does* remove one dependency, that being `remark-breaks` from the client package. I found that this plugin was responsible for the majority of strange behavior, and as far as I can tell:
1. It's intended behavior is already achieved by the sanitization script.
2. This plugin is not used anywhere else in the project.

If this is wrong though, I can revert it ofc. :)

</strike>

| Before | After |
| --- | --- |
| image | image |


sharkaccino:
ok i was *completely* wrong about the `remark-breaks` package, and infact, removing it ended up breaking single linebreaks so i've added it back :skull: as it turns out, the other changes i made ended up fixing the issues i thought were caused by that package. 

in any case, *now* everything should be fixed and working as intended :)

insertish (approved):

PR #346, opened 2025-02-27
feat: outlined auth login fields

StupidRepo:

# before
before

# after
after


insertish:
I don't think we want to make this change here.

b1bradders:
😔 fair

PR #353, opened 2025-03-26
refactor: use Lingui for i18n

insertish:
it's time for one of these again
i found this library during my uni group project and the DX is fantastic

- [x] Replace `t` with Lingui equivalents everywhere
- [x] Remove old dependencies
- [x] Integrate hot loading of catalogs

Closes https://github.com/revoltchat/frontend/issues/274


insertish:
😊
CleanShot 2025-03-26 at 16 21 35@2x

insertish:
CleanShot 2025-03-31 at 16 20 03@2x

PR #400, opened 2025-07-24
Add CompositionPicker component with GIF search, favourites

Asraye:
Added functionality to GIF picker component with search, featured, and favourites tabs using the Tenor API & localStorage support for favourites.

PS: Comments are added with ChatGPT as I was too lazy to write them manually but wanted to add sufficient commenting, excluding the TODO comments :) 

### DIFF
--- a/packages/client/components/ui/components/floating/CompositionPicker.tsx
+++ b/packages/client/components/ui/components/floating/CompositionPicker.tsx
@@ -1,5 +1,13 @@
 import { useFloating } from "solid-floating-ui";
-import { For, JSX, Ref, Show, createEffect, createSignal, on } from "solid-js";
+import {
+  For,
+  JSX,
+  Ref,
+  Show,
+  createEffect,
+  createSignal,
+  on,
+} from "solid-js";
 import { Portal } from "solid-js/web";
 import { Motion, Presence } from "solid-motionone";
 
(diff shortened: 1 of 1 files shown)


Zomatree:
Thank-you for the PR however any Tenor API requests shouldn't be done from the client but instead done via the revolt api, this both allows the responses be cached and does not leak the api token, which should not be stored in the client.

Furthermore this should not be using localStorage to store user data but instead use the proper settings sync system to allow this data to be shared to other devices and clients.

There is not really much that can be salvaged from this PR after these points as this does not fit our plan for how we want to implement a gif section in the client.

In future it's worth messaging one of the members of the revolt dev team or sending a message in the Revolt Development server to try avoid situations like this.

ghost:
> Thank-you for the PR however any Tenor API requests shouldn't be done from the client but instead done via the revolt api, this both allows the responses be cached and does not leak the api token, which should not be stored in the client.
> 
> Furthermore this should not be using localStorage to store user data but instead use the proper settings sync system to allow this data to be shared to other devices and clients.
> 
> There is not really much that can be salvaged from this PR after these points as this does not fit our plan for how we want to implement a gif section in the client.
> 
> In future it's worth messaging one of the members of the revolt dev team or sending a message in the Revolt Development server to try avoid situations like this.

All good - I may then try and figure out the full self-hosting and implement it onto the API and everything, I hope it at least provides a general idea on how it be :P

insertish:
Closing PR out as per above.
Thank you for your time and effort though.

PR #412, opened 2025-08-19
fix: fixed an issue where quoted replies would not render markdown properly.

alexjyong:

Resolves: https://github.com/revoltchat/frontend/issues/411

Also made sure that more complicated replies (images, urls, codeblocks) didn't ruin formatting.

image

### DIFF
--- a/packages/client/components/markdown/index.tsx
+++ b/packages/client/components/markdown/index.tsx
@@ -101,12 +101,12 @@ const replyComponents = () => ({
   customEmoji: RenderCustomEmoji,
   mention: RenderMention,
   spoiler: RenderSpoiler,
-  
+
   strong: elements.strong,
   em: elements.emphasis,
   code: elements.code,
   del: elements.strikethrough,
-  
+
   p: (props: any) => <>{props.children}</>,
   h1: (props: any) => <>{props.children}</>,
   h2: (props: any) => <>{props.children}</>,


packages/client/components/ui/components/features/messaging/elements/MessageReply.tsx
```diff
@@ -107,6 +107,22 @@ const Link = styled("a", {
...
+      .replace(/```[\s\S]*?```/g, '[code block]')
+      .replace(/^>\s+/gm, '')
+      .replace(/^#{1,6}\s+/gm, '')
+      .trim();
+  };
+
+  const getDisplayContent = (content: string) => {
+    const stripped = stripComplexMarkdown(content);
+    return stripped || '[complex content]';
+  };
```
insertish: I don't think this solution will necessarily scale or respond well to future changes, and may have hidden caveats.
Just glancing through the regexes, I assume this doesn't handle things like math for example?

alexjyong: Yeah my concern with a more broad solution was it rendering too much and breaking how replies looked. I'll look into your proposed solution.

alexjyong:
image
@insertish Added in the requested changes. 

I noticed you mentioned math rendering, which this doesn't handle, but I noticed that in production, there isn't support for LaTeX math syntax (which I presumed you meant) so I figured it was out of scope for this PR change.

packages/client/components/markdown/index.tsx
```diff
@@ -94,6 +96,45 @@ const components = () => ({
...
+  img: Null,
+  video: Null,
+  figure: Null,
+  picture: Null,
+  source: Null,
+  audio: Null,
+  script: Null,
+  style: Null,
+  time: Null,
+  timestamp: Null,
```
insertish: I think the rest of these can probably just pass-through their text content where applicable (e.g. headings or lists) and timestamps are also okay to render.

e.g. suppose someone sends a message that renders "Event starts in 5 minutes", the reply would just display "Event starts in"; or they sent a message with a heading "Things we should do", and the reply wouldn't display the textual content of the heading which might be confusing

packages/client/components/ui/components/features/messaging/elements/MessageReply.tsx
```diff
@@ -107,6 +107,10 @@ const Link = styled("a", {
 export function MessageReply(props: Props) {
   const client = useClient();
 
+  const renderReplyContent = (content: string) => {
+    return renderSimpleMarkdown(content);
+  };
```
insertish: We should limit the length of the content to something reasonable like ~ 128 characters since everything longer than that likely won't appear anyways, I'm not sure what a good length would be.

insertish (changes requested):
Solution looks good :slightly_smiling_face: 
I think we can permit a bit more / restrict a bit more in some areas, added review comments

packages/client/components/ui/components/features/messaging/elements/MessageReply.tsx
alexjyong: Updated

packages/client/components/markdown/index.tsx
alexjyong: Updated!

alexjyong:
@insertish  added in the requested changes. Thanks for the feedback!
image

PR #413, opened 2025-08-21
feat: edit server role information

reeesespuffs:


packages/client/components/app/interface/settings/server/roles/ServerRoleEditor.tsx
```diff
@@ -1,20 +1,249 @@
...
+
+    if (editGroup.controls.colour.isDirty) {
+      changes.colour = editGroup.controls.colour.value ?? null;
+    }
+
+    const updatedRole = await props.context.editRole(props.roleId, changes);
+
+    editGroup.controls.name.setValue(updatedRole.name);
+    editGroup.controls.hoist.setValue(updatedRole.hoist || false);
+    editGroup.controls.colour.setValue(updatedRole.colour ?? null);
```
insertish: I don't think this should be necessary? Since we just sent these values and the fields should already have these values.

packages/client/components/app/interface/settings/server/roles/ServerRoleEditor.tsx
```diff
@@ -1,20 +1,249 @@
...
+              />
+            )}
+          </For>
+        </Row>
+        </Column>
+        </Row>
+        </Column>
+
+        <Column>
+          <Text>Hoist Role</Text>
```
insertish: Should be using `class="label"`

packages/client/components/ui/components/utils/Form2.tsx
```diff
@@ -128,8 +128,10 @@ const FormCheckbox = (
       <Checkbox
         {...remote}
         checked={local.control.value}
-        onChange={(value) => {
-          local.control.setValue(value);
+        onChange={(value: boolean | Event) => {
+          const checked = typeof value === "boolean"
+          ? value : (value.currentTarget as HTMLInputElement).checked
+          local.control.setValue(checked);
```
insertish: I missed this one, not sure what is going on here.
It should in theory only be one or the other?

insertish: I'm getting CustomEvent as the value, was there a scenario where this spat out a boolean?

reeesespuffs: i think the checkbox component expected a boolean instead of a CustomEvent and the checkbox wouldn't update correctly but if it works without it we should be able to remove it

insertish: Ah I see, my mistake, I think I screwed up the typings here.

insertish: I've updated the typings and removed this bit of code.

PR #419, opened 2025-08-25
FIX: Show settings cog only if user has permissions 

Asraye:
Haven't touched the main server settings cog as I'm not sure a way to do it which is nice (and would allow all people who'd actually need access to it to see it without looking and functioning horribly)

Sorry for the previous amount of commits, still figuring out how GitHub works and didn't realise I was able to edit an existing PR to add DCO, now figured out how to do it much nicer, so >-< apologises! learning in real time


packages/client/src/interface/navigation/channels/ServerSidebar.tsx
```diff
@@ -370,26 +370,25 @@ function Entry(
...
-                  config: "channel",
-                  context: props.channel,
-                });
-              }}
-            >
-              <MdSettings {...iconSize("14px")} />
-            </a>
-          </>
-        }
+    <Show when={props.channel.server?.havePermission("ManageChannel")}>
```
insertish, line 373: There are quite a few permissions that could allow access to channel settings, namely managing permissions or webhooks. (https://developers.revolt.chat/developers/api/permissions.html)

insertish (changes requested):
There's quite

ghost:
Meant to push the change to this, accidently opened a new PR (woops) 
https://github.com/revoltchat/frontend/pull/422

PR #420, opened 2025-08-25
FIX: Delete channel JSON error

Asraye:
I don't say this is a end-all solution as this issue seems to be related to the API, which might be better to fix there – But it is a temporary fix, also closes the modal upon channel deletion


insertish:
This should be fixed by https://github.com/revoltchat/api/pull/36

PR #422, opened 2025-08-26
FIX: Show settings cog only if user has permissions, and InviteOthers

Asraye:
Meant to push the change to this, accidently opened a new PR (woops), adds strings for settings cog, and also adds a check for invite permissions 

### DIFF
--- a/packages/client/src/interface/navigation/channels/ServerSidebar.tsx
+++ b/packages/client/src/interface/navigation/channels/ServerSidebar.tsx
@@ -326,29 +326,31 @@ function Entry(
 ) {
   const { openModal } = useModals();
 
-  // determine if user can edit channel (any relevant permission)
-  const canEditChannel = [
-    "ManageChannel",
-    "ManagePermissions",
-    "ManageWebhooks",
-  ].some((perm) => props.channel.server?.havePermission(perm));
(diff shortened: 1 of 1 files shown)


packages/client/src/interface/navigation/channels/ServerSidebar.tsx
```diff
@@ -325,6 +326,16 @@ function Entry(
...
+  // determine if user can edit channel (any relevant permission)
+  const canEditChannel = [
+    "ManageChannel",
+    "ManagePermissions",
+    "ManageWebhooks",
+  ].some((perm) => props.channel.server?.havePermission(perm));
+
+  // user can create invites??
+  const canInvite = props.channel.server?.havePermission("InviteOthers");
+
```
insertish: Solid.js components run once, so these should be in functions to be reactive afaik
i.e. `() => yourCheck` and then `canInvite()` where it is used

insertish (changes requested):

packages/client/src/interface/navigation/channels/ServerSidebar.tsx
ghost: Updated the channel permission and invite checks to use `createMemo`

PR #423, opened 2025-08-26
FEAT: Join Server Preview 

Asraye:
Current issues; 
- It doesn't like you trying to join servers you are already in, I couldn't seem to make it correctly skip and go to navigation for this.
- Joining a server will navigate you to the correct channel, but not switch the server sidebar to correctly fit

↑ I am unsure on how to fix these above issues, and so far, any more changes to try and fix them, breaks it further, I (will likely) continue to try and fix them, unless someone else wishes too.. which I wouldn't be opposed 

### DIFF
--- a/packages/client/components/modal/modals/JoinServer.tsx
+++ b/packages/client/components/modal/modals/JoinServer.tsx
@@ -1,10 +1,11 @@
+import { createSignal, onCleanup } from "solid-js";
 import { createFormControl, createFormGroup } from "solid-forms";
 
 import { Trans } from "@lingui-solid/solid/macro";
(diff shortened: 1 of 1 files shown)


insertish:
I think it would be more sensible to just re-use the invite preview modal, would be more DRY and provide a consistent flow for server invites. (it shows the background, etc as well)

ghost:
> I think it would be more sensible to just re-use the invite preview modal, would be more DRY and provide a consistent flow for server invites. (it shows the background, etc as well)

I've begun having a look at doing this, I'll update it once It's completed – current issue is making the invite modal accept group invites.

PR #452, opened 2025-08-31
fix: AgeGate being unreadable in darkmode & add revite's geoblock feature

reeesespuffs:

Closes https://github.com/revoltchat/frontend/issues/433

### DIFF
--- a/packages/client/src/interface/channels/AgeGate.tsx
+++ b/packages/client/src/interface/channels/AgeGate.tsx
@@ -1,13 +1,18 @@
-import { JSXElement, Match, Switch, createEffect } from "solid-js";
+import { JSXElement, Match, Switch, onMount, createSignal } from "solid-js";
 
 import { Trans } from "@lingui-solid/solid/macro";
 import { styled } from "styled-system/jsx";
 
 import { useState } from "@revolt/state";
 import { LAYOUT_SECTIONS } from "@revolt/state/stores/Layout";
-import { Button, Checkbox, iconSize } from "@revolt/ui";
+import { Button, Checkbox, iconSize, Text } from "@revolt/ui";
 
 import MdWarning from "@material-design-icons/svg/round/warning.svg?component-solid";
+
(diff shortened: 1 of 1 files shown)


packages/client/src/interface/channels/AgeGate.tsx
```diff
@@ -21,21 +26,55 @@ export function AgeGate(props: {
...
+    fetch("https://geo.revolt.chat")
+      .then((res) => res.json())
+      .then((data) => setGeoBlock(data))
+      .catch(() => setGeoBlock(null));
+  });
+
   return (
     <Switch fallback={props.children}>
-      <Match when={props.enabled && (!confirmed || !allowed)}>
+      <Match when={geoBlock() && geoBlock()!.isAgeRestrictedGeo && props.enabled}>
```
insertish: I don't think we can risk permitting access if geoblocking service is momentarily unresponsive or is taking a moment to load

packages/client/src/interface/channels/AgeGate.tsx
```diff
@@ -21,21 +26,55 @@ export function AgeGate(props: {
...
   );
 
+  const [geoBlock, setGeoBlock] = createSignal<GeoBlock | null>(null);
+
+  onMount(() => {
+    fetch("https://geo.revolt.chat")
+      .then((res) => res.json())
+      .then((data) => setGeoBlock(data))
+      .catch(() => setGeoBlock(null));
+  });
```
insertish: This won't cache between channels, should use Tanstack `useQuery` instead, can use `<Suspense fallback={<loading icon>} />` therefore (there's at least one example of this in the code or refer to https://tanstack.com/query/latest/docs/framework/solid/overview#enough-talk-show-me-some-code-already

Query key would be something like `geoblock`

insertish (changes requested):

reeesespuffs:
I have updated the geoblock  to use `useQuery`

PR #454, opened 2025-09-01
feat: add "Copy User ID" to UserMenu

Asraye:
Adds a "Copy User ID" selector on the user menu for ease of access
image

### DIFF
--- a/packages/client/src/interface/navigation/servers/UserMenu.tsx
+++ b/packages/client/src/interface/navigation/servers/UserMenu.tsx
@@ -224,6 +224,13 @@ export function UserMenu(props: Props) {
                   <Trans>Clear status</Trans>
                 </ContextMenuButton>
               </Show>
+
+              <ContextMenuButton
+                icon={MdInfo}
+                onClick={() => navigator.clipboard.writeText(user()!.id)}
+              >
+                <Trans>Copy User ID</Trans>
+              </ContextMenuButton>
             </ContextMenu>
           </Motion>
         </Show>


insertish:
I think this makes more sense to hide this away in settings (even at that, because I've seen some people get confused on whether that's what they share to get others to add them)?

To non developers, this is probably confusing
NB. copy ID buttons across the app should probably be gated now that I think about it

ghost:
> I think this makes more sense to hide this away in settings (even at that, because I've seen some people get confused on whether that's what they share to get others to add them)?
> 
> To non developers, this is probably confusing NB. copy ID buttons across the app should probably be gated now that I think about it

Could pull a Discord and hide it under developer mode (lol), Or we could make it similar to Revite where it's in the settings menu as you suggested — but I can confirm a lot of people struggled to realise that was just a UserID and not something sensitive. I think having it in the status menu makes it clear what it actually is. we could also add an info hover next to it explaining what it is and that It's not sensitive as well to make it more clear? Whether we went with the status or settings approach

insertish:
Okay this is fine now actually, I added a settings toggle for IDs

insertish (changes requested):
Gate button behind advanced setting for IDs (see other context menus)

ghost:
> Gate button behind advanced setting for IDs (see other context menus)

Added – I've ran this through prettier as well this time, so.
Accidently removed the original commit with rebase, so it looks a bit weird now.

PR #465, opened 2025-09-16
feat: enhance error handling for file uploads and implement modal notifications for size limits

alexjyong:

Changed the alert message for too big file uploads to be a modal. Also updated the error handling modals to be able to take a custom error message.

image

image


packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -176,15 +187,39 @@ export function MessageComposition(props: Props) {
...
-    const validFiles = Array.from(files).filter(
-      (file) => file.size <= 20_000_000,
-    );
+    if (rejectedFiles.length > 0) {
+      const maxSizeFormatted = humanFileSize(MAX_FILE_SIZE);
+
+      if (rejectedFiles.length === 1) {
+        const file = rejectedFiles[0];
+        const fileSize = humanFileSize(file.size);
+        const error = new Error(`The file "${file.name}" (${fileSize}) exceeds the maximum size limit of ${maxSizeFormatted}.`);
```
insertish: These should be localised (see https://revoltchat.github.io/frontend/components/i18n/using-lingui.html)
Though note there is a mistake in the docs for the template string, it should use ${} as with any other template string.

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -176,15 +187,39 @@ export function MessageComposition(props: Props) {
...
+        error.name = "File too large";
+        openModal({
+          type: "error2",
+          error,
+        });
+      } else {
+        const error = new Error(`${rejectedFiles.length} files exceed the maximum size limit of ${maxSizeFormatted} and were not uploaded.`);
```
insertish: Same here

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -26,12 +27,21 @@ import {
...
 } from "@revolt/ui";
 import { generateSearchSpaceFrom } from "@revolt/ui/components/utils/autoComplete";
 
 import MdEmoji from "@material-design-icons/svg/filled/emoji_emotions.svg?component-solid";
 import MdSend from "@material-design-icons/svg/filled/send.svg?component-solid";
 
+/**
+ * Max file size allowed for uploads (in bytes)
+ * 20 MB = 20 * 1024 * 1024 = 20,971,520 bytes
+ * I kinda wonder if this should be a setting, or something fetched from the backend dynamically.
```
insertish: This is fine for now, it will be in the future, though if it's not much of a bother, you could move this into `env.ts` where `CONFIGURATION` lives so it's not missed later

alexjyong: @insertish Sure thing. Related to this, I was doing a related thing on my Android fork, kinda wondering if there is a way to query the file size limit from the backend server?  

Could be useful for Revolt clients not using this codebase, especially if they are interfacing with Revolt instances with different file limits.

packages/client/src/interface/channels/text/Composition.tsx
alexjyong: Resolved.

packages/client/src/interface/channels/text/Composition.tsx
alexjyong: Resolved.

alexjyong:
@insertish Updated the code per your feedback. Still working:
image

Edit: fixed nvm.
~~I notice that we have failing checks now. Haven't dug into them, but seems to be related to this commit? https://github.com/revoltchat/frontend/commit/8be60c398e1a52ec234fc267a8225f2f309c7220#diff-400d5902c167d2e00c9f4e5fee8aae5880f4f2edc6cd8ff6d0c31be204d731cb~~

packages/client/src/interface/channels/text/Composition.tsx
insertish: > @insertish Sure thing. Related to this, I was doing a related thing on my Android fork, kinda wondering if there is a way to query the file size limit from the backend server?
> 
> Could be useful for Revolt clients not using this codebase, especially if they are interfacing with Revolt instances with different file limits.

There needs to be some changes to the API to expose more information.

packages/client/src/interface/channels/text/Composition.tsx
alexjyong: Resolved

PR #474, opened 2025-09-25
fix: preserve replies when sending GIFs

pparaxan:
I'm still playing around w/ TypeScript but atleast I got to fix a Revolt bug lmao.

---


insertish (changes requested):
This has the side-effect of also sending any attached files and wiping the current draft which the user may intend to keep writing.

I think this logic should only take replies if there is no other content (text or files) attached to the draft.

PR #476, opened 2025-10-01
fix: wrap long links and fix underline in LinkWarning modal

Aeledfyr:
Before:
Screenshot From 2025-09-30 22-44-54
After:
Screenshot From 2025-09-30 23-28-59

The existing underlines styles weren't taking effect -- there's some bug in the interaction between the `<Trans>` component and inner `styled` components.  This PR works around that by moving the `<Link>` components outside of the `<Trans>`, but that has the side-effect of leaving trailing spaces in the `<Trans>` component and separating the `?`.  I'm not sure how to manage that better -- if you know how, feel free to edit this PR, or just close it and fix it separately.


insertish:
This seems fine as there's another piece of text doing something similar below, I just removed the question mark translation.

In the future, ideally the issue with Lingui here would be fixed.

PR #479, opened 2025-10-02
File/Attachment context menu

amycatgirl:

## Brief overview of the changes
- Added a new context menu entry, FileContextMenu (filename differs from the component name, should be renamed)

Nobody filed an issue, so this PR doesn't close anything.


packages/client/components/app/menus/AttachmentContextMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <ContextMenu>
+      <ContextMenuButton icon={MdOpenInNew} onClick={OpenFile}>
+        <Trans>Open file</Trans>
+      </ContextMenuButton>
+      <ContextMenuButton icon={MdLink} onClick={CopyLink}>
+        <Trans>Copy link</Trans>
+      </ContextMenuButton>
+      <ContextMenuButton icon={MdDownload} onClick={SaveFile}>
+        <Trans>Save file</Trans>
+      </ContextMenuButton>
```
insertish: Open file should open the originalUrl as users are likely not intending to open the preview
Save file should, instead of using onClick, be wrapped in an anchor with the `download` property set

i.e. match behaviour in ImageViewer.tsx (NB. the ImageViewer buttons don't include a share button but it could be added there)

amycatgirl: Resolved in ed7c9a4

PR #484, opened 2025-10-04
Add Mention permissions

amycatgirl:
Revolt.js already supports MentionEveryone and MentionRoles permissions, but they weren't added to the permission editor.

Now they are

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3387,10 +3387,10 @@ packages:
       react-dom:
         optional: true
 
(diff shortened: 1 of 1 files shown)


amycatgirl:
Adding autocomplete entries for roles and (at)everyone/(at)silent would also be nice. But I think that would warrant a separate PR

insertish:
Tried to undo changes but ended up deleting the file entirely

insertish:
According to my local project, Git says the lock files are the same, why is GitHub marking them as different ????

PR #495, opened 2025-10-15
fix: prevent adding more replies than max replies

reeesespuffs:

- [x] I understand and have followed the [contribution guide](https://developers.stoat.chat/contrib.html)


reeesespuffs:
> This should prevent adding more replies instead of just silently failing to send.

This prevents adding more replies

insertish:
Oops! Assumed this was in the send routine, LGTM!

PR #500, opened 2025-10-21
fix: allow non-custom emojis as reactions

Aeledfyr:
This fixes the bug preventing unicode emojis from being used as reactions -- specifically, the ProseMirror widgets for emojis now uniformly use id to represent the emoji string, rather than having unicode emojis have it as their textContent.

This PR will be obsoleted if/when the emoji picker results are no longer represented directly using ProseMirror widgets, which will likely need to happen for the CodeMirror-based editor.


insertish (approved):

insertish:
Superseded by text editor rewrite.

PR #504, opened 2025-10-30
feat: reimplement adding bots via discover

amycatgirl:
A few users in the main stoat server were asking in support how to add bots via discover because they were getting a missing handler error. So I decided to implement the missing handler. This code only adds relevant logic for matching bot links and hooking it up to the discover wrapper.

The code used to display the `add_bot` modal was taken verbatim from the development page, but it works (tested by mocking the discover frontend manually). However, it does not redirect the user to the server they added the bot to.


insertish (approved):
Adding bots is multi select so I wouldn't expect it to redirect, maybe it could if there's only one 🤔 but this is fine!

insertish:
Oh, it seems to be missing an import (see CI fail)

PR #515, opened 2025-11-13
fix: reset profile bio from latest data

### LINKED ISSUE #513: bug: Bio resets visually after save
When editing bio and then saving, the bio text field reverts state to the previous one and not the new updated one.

Probably caused by profile.data getting reset without getting new data beforehand

rexxDigital:
Fixes #513

Not sure if refetching is the preferred approach here, happy to change it if you’d rather handle this differently.

Edit:
We can just not reset from stale data of course

### DIFF
--- a/packages/client/components/app/interface/settings/user/profile/UserProfileEditor.tsx
+++ b/packages/client/components/app/interface/settings/user/profile/UserProfileEditor.tsx
@@ -75,7 +75,6 @@ export function UserProfileEditor(props: Props) {
 
   function onReset() {
     editGroup.controls.displayName.setValue(props.user.displayName);
-    // editGroup.controls.username.setValue(props.user.username);
     editGroup.controls.avatar.setValue(props.user.animatedAvatarURL);
 
     if (profile.data) {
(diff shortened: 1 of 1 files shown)


packages/client/components/app/interface/settings/user/profile/UserProfileEditor.tsx
```diff
@@ -135,12 +130,14 @@ export function UserProfileEditor(props: Props) {
...
+    await profile.refetch();
+    onReset();
   }
 
-  const submit = Form2.useSubmitHandler(editGroup, onSubmit, onReset);
+  const submit = Form2.useSubmitHandler(editGroup, onSubmit);
 
   return (
-    <form onSubmit={submit}>
+    <form onSubmit={submit} onReset={onReset}>
```
insertish: I don't believe this does anything

packages/client/components/app/interface/settings/user/profile/UserProfileEditor.tsx
```diff
@@ -135,12 +130,14 @@ export function UserProfileEditor(props: Props) {
     }
 
     await props.user.edit(changes);
+    await profile.refetch();
+    onReset();
```
insertish: onReset is already called by the submit handler after this function body exits

packages/client/components/app/interface/settings/user/profile/UserProfileEditor.tsx
```diff
@@ -135,12 +130,14 @@ export function UserProfileEditor(props: Props) {
     }
 
     await props.user.edit(changes);
+    await profile.refetch();
```
insertish: We should instead just update the queryClient's value of the profile.

insertish:
Refactored onReset back in to make this a little bit cleaner, see 18bd7d0f74839cec55973ccaff8ec3753dac8384.

Thanks for the contribution!

PR #517, opened 2025-11-14
fix: truncate UserMenu status text

zrodevkaan:
Truncates the UserMenu ContextMenu status text.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+  pull_request:
 
 jobs:
   deploy:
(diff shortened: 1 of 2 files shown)


packages/client/src/interface/navigation/servers/UserMenu.tsx
```diff
@@ -37,6 +37,16 @@ interface Props {
   anchor: Accessor<HTMLDivElement | undefined>;
 }
 
+const TruncatedStatusText = styled("div", {
+  base:
+  {
+    maxWidth: '300px',
```
insertish: Little bit arbitrary, but looks fine to me.

insertish (approved):

packages/client/src/interface/navigation/servers/UserMenu.tsx
zrodevkaan: Should I use a var that you guys approve of?

insertish: I suppose this could be a variable in layout: https://github.com/stoatchat/for-web/blob/f6a0377227519428ef560e82a45e3d69265d4db1/packages/client/components/ui/themes/stoatWebTheme.ts#L95

insertish:
Heads-up DCO failing

PR #538, opened 2025-11-21
fix: Report Content Message overflows with embeds. 

pridgey:
The fix here is with Report Content. Messages with embeds overflow the 240px, so instead I've changed it to an 80vh and an overflowY of hidden.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/.github/workflows/release-webhook.yml
+++ b/.github/workflows/release-webhook.yml
@@ -1,0 +1,19 @@
+name: Release Webhook
+
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+jobs:
+  release-webhook:
+    name: Send Release Webhook
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Send release notification webhook
+        run: |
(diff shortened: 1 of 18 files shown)


reeesespuffs:
Please correct me if im wrong. this looks like 2 bug fixes in one. please try to make a separate PR for each fix

pridgey:
> Please correct me if im wrong. this looks like 2 bug fixes in one. please try to make a separate PR for each fix

Fixing one without the other kind of felt like fixing a UI issue while the usability was still broken. Because it wasn't too big a change, I thought it would be okay to include in a single PR since it was all in the same screen, but I'd be happy to separate it if you think that would be best.

reeesespuffs:
considering the FloatingSelect feature doesn't appear to be used in the ReportContent modal it would indeed be best to move it into its own PR

pridgey:
> considering the FloatingSelect feature doesn't appear to be used in the ReportContent modal it would indeed be best to move it into its own PR

Okey doke, removed!

reeesespuffs:
~~one last thing before i can merge your PR :D DCO is failing, please remember to sign off your commits~~
please disregard the above, i have set DCO to pass

PR #546, opened 2025-11-23
feat!: profile picture roundness

### LINKED ISSUE #377: [OP#170] Select avatar roundness
**[View issue on OpenProject](https://op.revolt.wtf/projects/revolt-for-web/work_packages/170/activity)**

amycatgirl:
should close #377 if everything goes according to plan

Icons come from the stoatchat/for-android repository (converted from Android vectorDrawable to svg)

This adds a new theme option, `customRadii`, that tells the avatar component to change it's roundness (as a percentage). It also removes the "shape" prop, as i don't think will be needed anymore (for now). To override the roundness of the avatar, use the "roundness" prop instead.

Very breaking PR, btw

### DIFF
--- a/packages/client/components/ui/components/features/voice/VoiceChannelPreview.tsx
+++ b/packages/client/components/ui/components/features/voice/VoiceChannelPreview.tsx
@@ -152,11 +152,7 @@ function CommonUser(props: {
       }}
     >
       <Ripple />
-      <Avatar
-        size={24}
-        src={user().avatar}
-        fallback={user().username}
-      />{" "}
+      <Avatar size={24} src={user().avatar} fallback={user().username} />{" "}
       <PreviewUsername>{user().username}</PreviewUsername>
       <Row gap="sm">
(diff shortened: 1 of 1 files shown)


reeesespuffs:
Is there a reason this PR is marked as a Draft?

amycatgirl:
I marked this as a draft since I still want to do some things before marking it as ready for review, but the changes in the PR work.

insertish:
How breaking is this? Are we just talking internal APIs or external use? Because this would trigger a 1.x.x release.

amycatgirl:
Sorry, I have been pretty busy these past couple of days so I couldn't respond right away.

IIRC, I marked it as breaking because I changed the name of a prop in the avatar component which might break other pull requests. It also makes the avatar component dependent on the state context, which might have caused a performance regression (I haven't touched this branch in a while, so i am not sure if that's truly the case).

Not sure if it qualifies as a breaking PR though

amycatgirl:
superseded by #601, closing

PR #550, opened 2025-11-28
fix: codemirror bounding box not taking full composer height

amycatgirl:
Another issue that nobody reported until now, apparently.

As far as I know, TextEditor2 is used in about 2 or 3 places currently, i've checked them all and it seems to work just fine.

### DIFF
--- a/packages/client/components/ui/components/features/texteditor/TextEditor2.tsx
+++ b/packages/client/components/ui/components/features/texteditor/TextEditor2.tsx
@@ -245,6 +245,7 @@ export function TextEditor2(props: Props) {
 const editor = css({
   display: "inline-flex",
   height: "100%",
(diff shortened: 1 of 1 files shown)


reeesespuffs:
This seems to break being able to refocus the message box when editing a message. as well as making the message box placeholder unaligned

amycatgirl:
Fixed, though im not sure about the placeholder being misaligned. I pulled the changes locally and the placeholder looks properly aligned, at least it does seem like it on firefox. I don't have chrome or a webkit-based browser installed right now, so I can't check if that's the case in any of those.

image
image

reeesespuffs:
oh seems like the placeholder being misaligned was a mistake on my part, not sure what was causing that

PR #553, opened 2025-12-04
feat(messaging): animate profile picture on hover

amycatgirl:
Small little parity feature that adds back the ability to see animated profile pictures on messages. If and when we add accessibility settings to the app, this should be dictated by the user's `reduce-motion` preference.

Demo
---
https://github.com/user-attachments/assets/523430dd-4bad-45fa-848b-3d2f20286bc3


amycatgirl:
and i forgot DCO again, great

PR #568, opened 2025-12-20
fix: make reset token button functional

amycatgirl:
Mimics the behaviour found in revite (prompt via modal, then reset)


reeesespuffs:
Heads up, Lint & Format check is failing. please remember to format

PR #578, opened 2025-12-30
fix: Pass room to Tracks in RoomAudioManager

### LINKED ISSUE #577: Voice: No room provided
### What happened?

Quite suddenly I am no longer able to join any voice call. Neither via the browser nor via the desktop app.
According to the browser logs the error is "No room provided, make sure you are inside a Room context or pass the room explicitly". It stays on connecting, it also incorrectly shows me as still in the call when I clicked disconnect.
My tracks get published so other participants can hear me, I cannot hear them. Leaving the room shows the room, roomID etc rejoining shows them all as null. Rejoining does not show the no room provided error message but whilst it connects and others hear me, I cannot hear them and the callcard also doesn't shown me or the other participants

matthias-hoste:
This PR fixes #577 (I had no permission to reopen it and was being too fast). Sometimes(only me as far as I know) something causes the room in Tracks to break causing a fail to join. This is fixed by passing it

### DIFF
--- a/packages/client/components/rtc/components/RoomAudioManager.tsx
+++ b/packages/client/components/rtc/components/RoomAudioManager.tsx
@@ -1,5 +1,5 @@
 import { createEffect, createMemo } from "solid-js";
-import { AudioTrack, useTracks } from "solid-livekit-components";
+import { AudioTrack, useTracks, useMaybeRoomContext } from "solid-livekit-components";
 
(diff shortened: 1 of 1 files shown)


reeesespuffs:
Heads up, DCO failing. Lint & Format is also failing, please remember to format

Dadadah:
This PR doesn't actually change anything as the useTracks hook calls useEnsureRoom, which fetches the room from the context using the `useMaybeRoomContext` call.

PR #585, opened 2026-01-14
feat: Add a FloatingSelect Component

pridgey:
Adds a FloatingSelect component to be used in modals as the modal's dropdown is competing for layering.

I noticed @insertish had some quick scribbles with the FloatingSelect and I like the concept. Fixed some things so it could target the mdui-meni-items in the shadow dom and successfully grab their values.

image


insertish:
I'm not sure what happened here but this seems to have been superseded by #819.
That code was never meant to go to main, but I guess #819 got merged down.
It'll probably be replaced in due time.

PR #595, opened 2026-01-24
feat: prioritize displayname over username on messages

reeesespuffs:


insertish:
I realise now that there is a lot of hidden behaviour in stoat.js that should probably be avoided, being explicit like this seems way more maintainable (esp. since it's really just a few places where this applies -- two(?))

PR #619, opened 2026-02-10
chore: properly localize authifier errors

amycatgirl:
I think i need feedback on how some of these are worded.

But this PR adds translation strings for some common authifier errors, so people don't come into the support channel asking why they suddenly can't log into their accounts when they've entered their password wrong a bunch of times.

### DIFF
--- a/packages/client/components/i18n/errors.ts
+++ b/packages/client/components/i18n/errors.ts
@@ -111,7 +111,7 @@ export function useError() {
         case "LockedOut":
           return t`You have been locked out for entering a wrong password multiple times. Please wait a couple minutes and try again.`;
         case "CompromisedPassword":
-          return t`This password has been compromised before, please use another password.`;
+          return t`This password has previously appeared in security leaks, please use another password.`;
         case "UnverifiedAccount":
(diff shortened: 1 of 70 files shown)


insertish (approved):

insertish:
These are fine, I'll just throw past Jen to double check

packages/client/components/i18n/errors.ts
```diff
@@ -99,6 +106,16 @@ export function useError() {
           return t`You've sent too many friend requests, the maximum is ${err.max}`;
         case "PayloadTooLarge":
           return t`Your message is too long, please remove some characters and try again.`;
+        case "ShortPassword":
+          return t`The password is too short.`;
+        case "LockedOut":
+          return t`You have been locked out for entering a wrong password multiple times. Please wait a couple minutes and try again.`;
+        case "CompromisedPassword":
+          return t`This password has been compromised before, please use another password.`;
```
infi: should be: This password has previously appeared in security leaks, please use another password.

infi (changes requested):

insertish (approved):

PR #646, opened 2026-02-11
feat: custom css

amycatgirl:
initial draft for the custom css feature

---

wishlist:
- [ ] detachable editor
- [ ] debounced live-reload


kasivulpis:
I just wanted to ask how stable/usable is this right now? Asking because I wanted to pull your changes in to my docker image for my personal instance. Great work looking forward to this getting merged.

amycatgirl:
The editor works quite well, however it is quite useless without elements having "stable"css classes, as they could change at any point after recompiling.

kasivulpis:
> "stable" css classes

I'm sure you already know but that's just the minifier, which means I should be able to turn off the random classes and pull in your changes. Tysm for your work I'll try your changes out when I get the chance! I wonder how you'll handle this in the future though, maybe we have use IDs on specific containers/elements and the css cascades from there? I think that's probably the best idea but IDK.

amycatgirl:
We do not use any CSS minifiers.

Instead, we use a CSS-in-JS solution called PandaCSS, which generates CSS classes on-the-fly to style components such as containers, buttons, and other custom elements. By "stable", I mean a proper class name for each atom (for instance, Buttons would be `.button`, sliders `.slider`, and so on).

It will be done in a separate pull request stacked on top of this one.

kasivulpis:
Oh I see! Sorry I don't know your techstack (or i guess toolchain?) so I assumed that it was an issue with all the class names being minified like I've seen in other prod builds I see. Ty for clarifying.

insertish:
Please drop `i18n/catalogs` from the PR, workflow has been changed, sorry for the inconvenience :slightly_smiling_face: 

---

> wishlist:
> - [ ] detachable editor
> - [ ] get element classes and id without opening devtools
> - [ ] debounced live-reload

I think just having an editor (and de-bouncing the reload, I think we already have lodash.debounce available? or something equivalent at least) would be a good scope box.

amycatgirl:
I'll be re-scoping this PR. Closing it.

If anyone wants to take over, you can come to the stoat development server and discuss this.

PR #663, opened 2026-02-12
fix: border radius on connected button groups

penguinencounter:
Hopefully you don't mind this PR, apologies in advance if I did it wrong! The inconsistency was bothering me.

Before:
image

After:
image

roughly copied from the commit message:

> The current styling causes border radii on connected groups to be inconsistent between buttons in the middle vs buttons on the end of the group. (This is most obvious in the Appearance tab in the settings.)
> 
> This is due to how the CSS spec handles `border-radius` values that cause overlapping corners (ex. the ones used to create pill-shaped buttons here) - instead of only reducing the conflicting corners, all corners are reduced proportionally as needed:
> https://drafts.csswg.org/css-backgrounds/#corner-overlap
> 
> This is fixed by introducing a `--max-radius` CSS variable that is half of the button's height, which achieves the same effect without having corner overlaps. As a bonus, we reduce the `shape: "round"` variants to a single one that references the variable.
> 
> Note that there could potentially still be a consistency issue if a button is less wide than tall, but that seems very unlikely.

### DIFF
--- a/packages/client/components/ui/components/design/Button.tsx
+++ b/packages/client/components/ui/components/design/Button.tsx
@@ -281,22 +281,27 @@ const button = cva({
     size: {
       xs: {
         height: "32px",
+        "--max-radius": "16px",
         "--padding-inline": "12px",
       },
       sm: {
         height: "40px",
+        "--max-radius": "20px",
         "--padding-inline": "16px",
       },
       md: {
         height: "56px",
+        "--max-radius": "28px",
         "--padding-inline": "24px",
       },
       lg: {
         height: "96px",
+        "--max-radius": "48px",
         "--padding-inline": "48px",
       },
       xl: {
         height: "136px",
+        "--max-radius": "68px",
         "--padding-inline": "64px",
       },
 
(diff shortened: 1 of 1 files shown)


infi:
The affected buttons are Material 3 button groups.
Please ensure your changes comply with the spec for Material 3 button groups: https://m3.material.io/components/button-groups/specs

penguinencounter:
@infi they should, unless the existing `var(--borderRadius-[size])` values are not correct
design

Pecacheu:
Seems to comply with the material design guidelines to me. Make sure to drop the i18n files from this PR, though.

penguinencounter:
rebased! (hopefully the bot doesn't decide to change the lang again...)

Pecacheu (approved):

Dadadah:
Hi, looking at this again. I think this will pass UI review if you reduce the spacing between buttons in the button group to match the spec (2px, or --gap-xs)

Dadadah:
The border radii were entirely wrong, I've updated them to spec and pushed them. I've also added 2px gaps to each button group. Merging this.

Thanks for your contribution.

PR #697, opened 2026-02-13
feat: Adds a Docker Build for the Web Client

luthes:
This adds a multi-stage docker build for the web client. This was missing in the self-hosted compose (or at least was using the old UI).

* Multi stage Dockerfile - Builds a nice small alpine image with all the static output
* inject.js - Vite bakes env vars at build time, so I had to replace these (this also helps with multi-env without having a ton of different builds. One image, any environment). 
* Github Actions: CI for the docker builds. I just noticed there's no arm64 in the docker builds and I saw other builds with it, I can add it if needed.

Just a quick bit on the inject.js script, that runs at container startup and copies the built `dist/` to `dist_injected/`, does a find and replace of the placeholder strings with the actual env vars, and serves the result, but the original `dist` isn't touched, so containers restart cleanly. 

...Or do they. Now that I write that, I might have over-engineered that bit, and it only needs to do the replace, any container/pod restarts would just restore the original `dist/`. Let me know which direction you'd want to go with that, and I can modify it. The replace needs to happen regardless. Maybe it's useful to keep the original around :shrug: 

You can test it though:

* `docker build -t for-web .` 
* `docker run -e VITE_API_URL=https://api.stoat.chat -e VITE_WS_URL=wss://ws.stoat.chat -p 5000:5000 for-web`
* Verify env vars are injected in served JS

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ Before contributing, make yourself familiar with [our contribution guidelines](h
 
 Before getting started, you'll want to install:
 
-- Git
-- mise-en-place
+- [Git](https://git-scm.com/install/)
+- [mise-en-place](https://mise.jdx.dev/getting-started.html)
 
 Then proceed to setup:
 
--- a/packages/client/.env.example
+++ b/packages/client/.env.example
@@ -3,6 +3,7 @@ VITE_API_URL=http://localhost:14702
 VITE_WS_URL=ws://localhost:14703
 VITE_MEDIA_URL=http://localhost:14704
 VITE_PROXY_URL=http://localhost:14705
+VITE_GIFBOX_URL=http://localhost:14706
 
 # specify hCaptcha sitekey
 VITE_HCAPTCHA_SITEKEY=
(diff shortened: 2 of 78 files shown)


luthes:
I was just reading over the other Docker PR. I think the dist_injected directory actually helps (solves?) the security issue brought up in [this comment](https://github.com/stoatchat/for-web/pull/522#issuecomment-3705910600), and [the one below it](https://github.com/stoatchat/for-web/pull/522#issuecomment-3707173328). Or at least is an implementation of the tmpfs that was talked about there.

I am not sure that either Nginx + SSI or the env inject approach are any more secure than the other. If an attacker gets access to the file system, both methods will have a writable filesystem they can access. I'm not a security expert, but it doesnt' seem much different if they get container access in one vs. the other.

Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+ENV VITE_PROXY_URL=__VITE_PROXY_URL__
+ENV VITE_HCAPTCHA_SITEKEY=__VITE_HCAPTCHA_SITEKEY__
+ENV BASE_PATH=/
+
+RUN pnpm --filter client exec vite build
+
+# ============================================
+# Stage 2: Minimal runtime image
+# ============================================
+FROM node:22-alpine
```
chrishultin: Any particular reason to use Alpine 22, rather than either the latest or an LTS version?

Rexogamer: I think this is _Node_ 22? Alpine is on 3.x afaik

chrishultin: I think you're right - that's totally a brainfart on my end

luthes: Yeah, node 22. I can bump that to 24, since it's the newest LTS

Dadadah (approved):

reeesespuffs (approved):

PR #698, opened 2026-02-13
feat: Add an invite code input to the registration page

### LINKED ISSUE #639: Client doesn't support invite-only mode
### What happened?

If the server is set to invite-only mode, it's impossible to register a user with the web client (registration form shows no field for invite; requests fail with 400).

Ideally, it should have UI to allow entering an invite code.

Dadadah:
Add an invite code input to the registration page. Uses a memo to determine if the invite code input should be shown. 

Requires stoatchat/javascript-client-sdk#127 due to the configuration currently not being loaded automatically by the javascript client sdk.

Resolves #639

### DIFF
--- a/packages/client/components/auth/src/flows/FlowCreate.tsx
+++ b/packages/client/components/auth/src/flows/FlowCreate.tsx
@@ -6,8 +6,9 @@ import { Button, Row, iconSize } from "@revolt/ui";
 
 import MdArrowBack from "@material-design-icons/svg/filled/arrow_back.svg?component-solid";
 
-import { useApi } from "../../../client";
+import { useApi, useClient } from "../../../client";
 
+import { Show, createMemo } from "solid-js";
 import { FlowTitle } from "./Flow";
 import { setFlowCheckEmail } from "./FlowCheck";
 import { Fields, Form } from "./Form";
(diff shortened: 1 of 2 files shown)


packages/client/components/auth/src/flows/FlowCreate.tsx
```diff
@@ -27,11 +29,13 @@ export default function FlowCreate() {
     const email = data.get("email") as string;
     const password = data.get("password") as string;
     const captcha = data.get("captcha") as string;
+    const invite = data.get("invite") as string;
```
seva-luchianov, line 32: Could I propose looking for an `?invite=code` qs and setting `invite` to that value if `data.get("invite")` returns an empty string? It would make the signup flow for invited users a lot more straightforward.

Alternatively the qs code could prefill the invite field.

Either way this results in the user being registered not having to know anything about invite only functionality and/or copy pasting codes, but also gives them the flexibility to easily override the invite code from the qs with whatever they want if necessary.

Dadadah: That's a good idea. However, it's unlikely this pr will get merged, please leave the same comment on #769

seva-luchianov: Curious why this isn't getting merged? I assume the other PR is better bc no client dep?

Dadadah: The other PR has better flow and better UI. I'd rather it depend on this dependency, especially since this is for self hosted. If the dependency gets merged I may speak with the author of that pr to implement it.

PR #706, opened 2026-02-13
fix: Prioritize masquerade name on displayed username

### LINKED ISSUE #694: Masquerade name not rendered in v0.2.0
### What happened?

Currently a the displayname will be displayed on a post instead of the name from the masquerade block on a masqueraded post.
I suspect it's related to the changes to make the displayname render to posts instead of the username.

enoua5:
Fixes #694


danikr0:
Nice, we just found this bug and wanted to report it, but there already is the fix! If I could, I'd approve! :D

insertish (approved):

insertish:
checks failing due to unrelated changes

Issue #717, opened 2026-02-13
Message timestamp wrapping creating extra space below message


jamac-dumas:
### What happened?

The timestamp to the left of a message can wrap to two lines, leaving an awkward space below the message if the message is only one line.

<img width="429" height="356" alt="Image" src="https://github.com/user-attachments/assets/83c2729f-4428-457b-b672-55825ea92dff" />

<img width="493" height="128" alt="Image" src="https://github.com/user-attachments/assets/6dcd6502-b97d-476e-a0c7-8dcbdbb94bae" />

<img width="538" height="126" alt="Image" src="https://github.com/user-attachments/assets/bc60b839-ba47-41aa-8821-b55c15e8a3cc" />

foopis23:
Okay, i wrote a script to test this. I found that none compact variant of the info component wraps the timestamp for 151 different times at 52px, 18 times on 53px, and 0 times on 54px.

I'm putting together the pr to switch it to 54px now. 

Here is that script i used to test. Just replace the xpath with a valid xpath of a time component on the page. 
```js
const xpath = '/html/body/div[4]/div/div/div[2]/div/div[2]/main/div[2]/div/div/div[2]/div/div[31]/div[2]/div[1]/div/time';

const timeEl = document.evaluate(
  xpath,
  document,
  null,
  XPathResult.FIRST_ORDERED_NODE_TYPE,
  null
).singleNodeValue;

if (!timeEl) {
  console.error("Time element not found");
} else {
  const originalText = timeEl.textContent;
  const originalHeight = timeEl.getBoundingClientRect().height;

  const periods = ["AM", "PM"];
  const tallerTimes = [];

  for (let hour = 1; hour <= 12; hour++) {
    for (let minute = 0; minute < 60; minute++) {
      for (const period of periods) {
        const formatted =
          `${hour}:${minute.toString().padStart(2, "0")} ${period}`;

        timeEl.textContent = formatted;
        void timeEl.offsetHeight; // force layout

        const newHeight = timeEl.getBoundingClientRect().height;

        if (newHeight > originalHeight) {
          tallerTimes.push({
            time: formatted,
            height: newHeight
          });
        }
      }
    }
  }

  timeEl.textContent = originalText;

  if (tallerTimes.length === 0) {
    console.log("Element never grew taller.");
  } else {
    console.log("Height increased for:", tallerTimes);
  }
}
```

For those who are interested, here are the 18 times that overflow on 53px. The "widest" times if you will.
```json


[
[...]

PR #719, opened 2026-02-13
fix: info timestamp wrapping causing additional space below messages

foopis23:
fix: info timestamp wrapping causing additional space below messages

fixes issue #717


insertish:
I'm not sure why it auto requested (maybe the branch update?)

Issue #710, opened 2026-02-13
Validate names in the frontend


MHanak1:
Currently, values such as emote names or channel names are not validated on the client, which results in a generic error message after the request is rejected by the server.

The frontend should validate names with the same regex as the server, and provide feedback to the user if the value is invalid. If possible, the input field should display errors in real time.

Currently, i am aware of those fields which do not check the values on the client (there is certainly more):
- [ ] Emote Name
- [ ] Server Name
- [ ] Channel Name

guizaa:
Hello, I would like to make a PR for this issue to provide better feedback for the fields mentioned

MHanak1:
I've opened a pull request #801 which adds length limits to required text fields. It also adds missing props to the `TextField` component, one of them being the `pattern` prop, which can validate input fields based on regex.

Now to fully address this issue one would have to apply the `pattern` prop to the relevant fields, and add properly localised invalidity message (which i am unsure how to do)

PR #723, opened 2026-02-13
fix: sanitize emoji name on submit

mihaicm93:
sanitize emoji names client-side to match server validation (lowercase, digits, underscore)

### DIFF
--- a/packages/client/components/i18n/catalogs/ar/messages.po
+++ b/packages/client/components/i18n/catalogs/ar/messages.po
@@ -1169,6 +1169,10 @@ msgstr ""
 msgid "Emoji Pack (affects your messages only)"
 msgstr ""
 
+#: components/app/interface/settings/server/emojis/EmojiList.tsx
+msgid "Emoji Required"
+msgstr ""
+
 #: components/app/interface/settings/ServerSettings.tsx
 msgid "Emojis"
 msgstr "الرموز التعبيرية"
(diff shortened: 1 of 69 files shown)


mihaicm93:
https://github.com/stoatchat/for-web/issues/710

gabrielfordevelopment:
Please avoid hardcoding MIME types. Instead, store them in a list or configuration and iterate through it. This will make the code much easier to maintain in the future

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
```diff
@@ -34,9 +34,32 @@ export function EmojiList(props: { server: Server }) {
...
+          v
+            .toLowerCase()
+            .replace(/[^a-z0-9_]/g, "")
+            .slice(0, 32),
+      }),
       file: createFormControl<string | File[] | null>(null, {
         required: true,
+        validate: (files) => {
+          if (!files || files.length === 0) return t`Emoji Required`;
+          if (files[0].size > 500 * 1024)
```
chrishultin, line 49: We should probably use the actual values from the configuration, once that's available.

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
```diff
@@ -34,9 +34,32 @@ export function EmojiList(props: { server: Server }) {
...
+            return t`File size must be less than 500KB`;
+          if (
+            ![
+              "image/png",
+              "image/jpeg",
+              "image/jpg",
+              "image/gif",
+              "image/webp",
```
chrishultin, line 57: We need to make sure this is an exhaustive/accurate list, and that it isn't something that can be configured on the server

chrishultin (changes requested):

Issue #687, opened 2026-02-12
Generic error message for long channel names


Zekamon:
### What happened?

When creating a new channel with a name thats over 32 characters, a generic error message shows saying the following:

An error occured
Something went wrong! Try again later.

I assume a generic error message shouldn't show up here since it does not explain what the cause for the error is. Changing the channel name after creating automatically cuts off the characters above the limit without any error message.

PR #726, opened 2026-02-13
fix: limit channel name input to 32 chars

hyfetch:
Should fix https://github.com/stoatchat/for-web/issues/687

### DIFF
--- a/packages/client/components/modal/modals/CreateChannel.tsx
+++ b/packages/client/components/modal/modals/CreateChannel.tsx
@@ -19,14 +19,12 @@ export function CreateChannelModal(
   const { showError } = useModals();
 
   const group = createFormGroup({
-    name: createFormControl("", {required: true }),
+    name: createFormControl("", { required: true }),
     type: createFormControl("Text"),
   });
 
   async function onSubmit() {
-
     try {
-
       const channel = await props.server.createChannel({
         type: group.controls.type.value as "Text" | "Voice",
         name: group.controls.name.value,


packages/client/components/modal/modals/CreateChannel.tsx
```diff
@@ -19,7 +19,7 @@ export function CreateChannelModal(
   const { showError } = useModals();
 
   const group = createFormGroup({
-    name: createFormControl("", { required: true }),
+    name: createFormControl("", { required: true, maxLength: 32 }),
```
chrishultin: Why try to set this here, if it's not used in the check?

Also, I don't think that that is a valid parameter for that function.

chrishultin:
Can you also apply your fixes to packages/client/components/app/interface/settings/channel/Overview.tsx - ideally splitting the function out so that there's only one copy of the validation logic

mihaicm93:
I see this also touches a part from this Issue https://github.com/stoatchat/for-web/issues/710

MHanak1:
I believe this could also be done by passing the [maxlength](https://www.mdui.org/en/docs/2/components/text-field#attributes-maxlength) attribute to the underlying `mdui` component, which would probably be a cleaner implementation

MHanak1:
Soon I'll be opening a pull request which closes this as well as some other issues from #710 using the built-in `mdui` attributes

insertish:
Closing in favour of #710 
We shouldn't be trying to handle events for this kind of stuff.

PR #729, opened 2026-02-13
feat: Added Amoled Support, Fixed Border Radius Issues and made the contrast better for non-amoled Dark Mode

uaxdotnet:
Hi,

I fixed some UI issues with bad looking border radius being applied on the main UI and the settings page.

image
image

Didn't touch border radius for anything else right now since it looked alright but I'll work to make the UI uniform.

Additionally, removed the black background from the main chat window because the contrast was bad.

image

Edit #1 : 

Added Amoled Support
image

Thanks!


Pecacheu:
I think this is a preference thing, I mean personally I love the look of the rounded corners. Could be a setting?

Fully support AMOLED theme though.

uaxdotnet:
> I think this is a preference thing, I mean personally I love the look of the rounded corners. Could be a setting?

I've not removed most rounded corners. Just removed them for the settings screen(where it looked bad) and the main chat window where I fixed the contrast hence they're not visible anymore. They're available everywhere else right now.

Although it may be great to work on making the border radius variable as per the user's preference. In the future.

uaxdotnet:
I have decided to rework this PR to better comply with Material 3 design standards.

insertish:
Please split this PR into individual changes, the lingui workflow has also been updated now so please nuke the i18n folder changes in your PR.

PR #737, opened 2026-02-14
fix: workaround links that are actually intended for the app

amycatgirl:
was checking against https://stoat.chat without the /app to match current deployment

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# For-web maintainers
+* @stoatchat/for-web-maintainers
+
+# UI/UX related code
+/packages/client @stoatchat/ui-ux


amycatgirl:
of course, this should match the current deployment's current app link instead, it shouldn't be hardcoded.

which is why it's marked as a workaround.

Pecacheu:
Before, first link doesn't work, second link works.


After, first link works, second link no longer displays as a channel embed.


Unfortunately it appears the solution tis not so simple, in fixing one problem you created another. Maybe checking against `https://stoat.chat/` would work?

Edit: NVM there is a second problem, which is that we're checking this array against *url.origin*, which only includes the base URL (without any path) to begin with. Something like this might work then:
```js
[
  location.origin,
  // legacy
  "https://app.revolt.chat",
  "https://revolt.chat",
  // new
  "https://stoat.chat",
].includes(url.origin) &&
url.pathname.length > 1
```

amycatgirl:
today is not my day, huh

git is giving me issues, moment.

PR #740, opened 2026-02-14
feat: add context menu option for copying usernames

kasivulpis:
When trying to copy a username from the floating/modal user info elements, the entire banner is a giant button and you just open the modal/background banner respectively.

This means you need to write a user's username manually if you want to add them and aren't in the same server as them, or if you're trying to send your username to a friend on stoat who isn't in the same server as you.

image
image

### DIFF
--- a/packages/client/components/i18n/catalogs/pt-BR/messages.po
+++ b/packages/client/components/i18n/catalogs/pt-BR/messages.po
@@ -777,6 +777,10 @@ msgstr ""
 msgid "Copy user ID"
 msgstr ""
 
+#: components/app/menus/UserContextMenu.tsx
+msgid "Copy Username"
+msgstr ""
+
 #: components/app/interface/settings/channel/webhooks/ViewWebhook.tsx
 msgid "Copy webhook URL"
 msgstr ""
--- a/packages/client/components/i18n/catalogs/zh-Hans/messages.po
+++ b/packages/client/components/i18n/catalogs/zh-Hans/messages.po
@@ -777,6 +777,10 @@ msgstr ""
 msgid "Copy user ID"
 msgstr ""
 
+#: components/app/menus/UserContextMenu.tsx
+msgid "Copy Username"
+msgstr ""
+
 #: components/app/interface/settings/channel/webhooks/ViewWebhook.tsx
 msgid "Copy webhook URL"
 msgstr ""
--- a/packages/client/components/i18n/catalogs/zh-Hant/messages.po
+++ b/packages/client/components/i18n/catalogs/zh-Hant/messages.po
@@ -777,6 +777,10 @@ msgstr ""
 msgid "Copy user ID"
 msgstr ""
 
+#: components/app/menus/UserContextMenu.tsx
+msgid "Copy Username"
+msgstr ""
+
 #: components/app/interface/settings/channel/webhooks/ViewWebhook.tsx
 msgid "Copy webhook URL"
 msgstr ""


chrishultin (approved):

RatDotRaw:
To add extra functionality/feature.

I think it's also interesting to be able to highlight or just click to copy username (not display name). 
Similar to how discord used to do this as well.

Ofc **don't forget** some sort of feedback, so users know they copied something.

549754325-342d9bd8-ef38-44ed-8229-3cb26c6ddc96

infi:
(Moved to [review](https://github.com/stoatchat/for-web/pull/740#pullrequestreview-3940832996))

RatDotRaw:
@infi I don’t agree that copying should only live on username click.
Giving users multiple ways to do the same thing isn’t inherently bad or redundant.

Real-world example: websites. To get home you can click the logo, a “Home” link in the header, a footer link, sometimes even a “back to home” button on the page itself. None of that is considered a problem, it’s just accommodating different habits.

Same idea here. Clicking the username is fast, the context menu is explicit. They serve different users and don’t really conflict.

Just something to keep in mind ❤️

kasivulpis:
Hi sorry everyone, I've been too busy recently and have been neglecting my PRs. I was going to close this one to let someone more active fix it but like with #769 but I noticed this one is approved and just waiting for conflict resolution (aka just pulling the new lingui translations) and those two pipelines. Could someone explain what I'd need to do to fix those two pipelines so we can merge this in? Alternatively if someone else has already added a copy username feature I don't mind closing this PR too since I should probably only contribute again when I have time.

IAmTomahawkx:
You can ignore the two pipelines. They're not required to merge.

kasivulpis:
> This should happen when the username is clicked (with feedback in form of a temporary tooltip, ex. "Username copied"), in which case the context menu option is not necessary and should be removed

Yea that would be ideal but (when I made this) the component in the username popup is just a giant button and this would:
 1) Require someone on the team do decide what happens to the old banner functionality
   - Does it stay on the user popover banner click? If so will users be annoyed as hell when they click trying to auto copy the username and instead get the user modal?
   - If we're changing the banner functionality so that it does not open the user modal on click, then where is it going to go? New button? Different option/screen somewhere? Does the user popover now appear on hover and the user details modal appears on click instead?
 2) Is it 'obvious' enough that clicking the username will 'randomly' copy the username to clipboard from a ux perspective as apposed to:
   - A copy icon button next to the username? (Or even just a copy icon at all to indicate that its a thing you can do?)
   - A button/box that says 'Copy Username'.
   - Simply allowing users to select the text for a username and copy it manually.
   - P.S. The background banner randomly opening the user ui modal isn't the best ux either but thats outside of the scope of this comment/pr. 
 3) Last time I contributed/made this mr, the banner was *a giant button*. Assuming its still that way: Adding the ability to click on both the username specifically AND the original banner is asking to change how the user popover banner works and goes beyond the scope of this PR (OR we need to decide that the entire user popover banner now copies a username and the user details modal opens via some other way like I said in my first point.) If thats what we want that's totally fine, but thats outside the scope of this pr because thats another feature and has more work attached to it that is 'outside the scope' of this PR/change.

Also in case its not clear im not saying all of this to be frustrating on purpose or something. Maybe its because I'm used to my work environment and not foss contributions but it seems like 1-3 is obvious to me and I feel like I'm missing something. Especially my third point since this was a very strict rule for us at my old workplace and the workplaces of many of my cs friends. Usually asking a dev to 'upgrade' a existing feature/pr to something that starts escaping the scope of the original work is a massive no-no and you're supposed to create, groom, point, and then finally assign a new ticket to some dev instead (usually the same one.) I know obviously this isn't that kind of environment and there's no tickets, grooming, or pointing but asking this PR to go from:

> I made this tiny change that adds a context menu option that copies the username.

to

> I now need to figure out what to do with the old banner functionality and then either update the popover banner component to account for clicking in different regions, or to change its base functionality to copying usernames and then figure out where I'm moving access to the user details modal to, all while taking the ux in to consideration.

Is a bit much. Again I don't mean to come off as trying to be frustrating for no reason but its just weird for me from the environment I came from and I don't know if the disconnect is because the nature of FOSS contribution/github is just way different than gitlab/corporate, or if its because I'm stupid and theres some kind of other disconnect.

P.S. Like I said I'm busy now, so it's probably best to close this pr and add whatever debt/work is required for what I pointed out to wherever the 'help wanted' posts/feature requests go so someone else can work on it. If it really is as simple as 'just do what the maintainers say/want and it gets merged' in github/foss then my bad for over complicating things I'm just not really used to this sort of thing. 

Edit: Also ignore the pr review request, I clicked the wrong button my bad.

PR #751, opened 2026-02-14
fix: Updated README.md to include connecting to official backend

Jespercal:
I had trouble figuring out if it was possible to connect to the official backend when running a local instance of the web client.
It is possible, but it is not stated anywhere in the documentation or README.md, so I decided to help add it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ If you want the client to connect to the official hosted backend instead, open t
 
 ```
 
-When these variables are not set, the client automatically falls back to the official backend. (See https://github.com/stoatchat/for-web/blob/a113e2171224aa1938cd2b37793c43aa10397842/packages/client/components/common/lib/env.ts)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +70,22 @@ You can switch back to the fallback assets by running deinit and continuing as n
...
+```env
+# connect to local Stoat instance
+#VITE_API_URL=http://localhost:14702
+#VITE_WS_URL=ws://localhost:14703
+#VITE_MEDIA_URL=http://localhost:14704
+#VITE_PROXY_URL=http://localhost:14705
+
+```
+
+When these variables are not set, the client automatically falls back to the official backend. (See https://github.com/stoatchat/for-web/blob/a113e2171224aa1938cd2b37793c43aa10397842/packages/client/components/common/lib/env.ts)
```
chrishultin: Can we change this to point to `https://github.com/stoatchat/for-web/blob/main/packages/client/components/common/lib/env.ts` instead?  I worry that this link might go stale somehow.

Jespercal: You're right, I'll change that quickly.

PR #757, opened 2026-02-14
fix: Incorrect password field on account creation

SandPoot:
Don't be too cruel with me, i was just attempting to check out stoat and noticed that trying to create an account asks you to "Enter your current password" and doesn't have password suggesting.

I have replaced "password" with "new-password" according to the file Form.tsx, this is not AI-assisted, i've done it raw on my web browser.

### DIFF
--- a/packages/client/components/auth/src/flows/FlowCreate.tsx
+++ b/packages/client/components/auth/src/flows/FlowCreate.tsx
@@ -49,7 +49,7 @@ export default function FlowCreate() {
         <Trans>Hello!</Trans>
       </FlowTitle>
       <Form onSubmit={create} captcha={CONFIGURATION.HCAPTCHA_SITEKEY}>
-        <Fields fields={["email", "password"]} />
+        <Fields fields={["email", "new-password"]} />
         <Row justify>
           <a href="..">
             <Button variant="text">


packages/client/components/auth/src/flows/FlowCreate.tsx
```diff
@@ -49,7 +49,7 @@ export default function FlowCreate() {
         <Trans>Hello!</Trans>
       </FlowTitle>
       <Form onSubmit={create} captcha={CONFIGURATION.HCAPTCHA_SITEKEY}>
-        <Fields fields={["email", "password"]} />
+        <Fields fields={["email", "new-password"]} />
```
acidtib, line 52: this will break the api call with `422: Unprocessable Entity` since the request sends `password` not `new-password`

to complete this pr we would need something like 
L66-71
```
  // Map field types to form field names
  // "new-password" should submit as "password" for API compatibility
  const getFieldName = (field: Field) => {
    if (field === "new-password") return "password";
    return field;
  };
```

L85
```
              name={getFieldName(field)}
```
in `packages/client/components/auth/src/flows/Form.tsx`

acidtib (changes requested):
thank for the work

packages/client/components/auth/src/flows/FlowCreate.tsx
SandPoot: I'm sorry, i'll have to check this later, ideally, i wouldn't want stuff to be hacky like this, but if that's fine i'll just apply it

Dadadah:
This was fixed by #1052

PR #760, opened 2026-02-14
fix: Bypass email registration page if the server configuration has email disabled

### LINKED ISSUE #759: Email verification page is shown after account creation regardless of server configuration
### What happened?

Email verification page is shown after account creation regardless of whether the server has email verification enabled. There is a // TODO regarding it. The web app would have to fetch the server's config and skip the email verification step of the flow if email is disabled.

Dadadah:
After registration, the verify email page is shown regardless of whether email is enabled or not. This PR makes the account creation flow skip email verification and log straight in if the server is configured with email disabled.

Requires https://github.com/stoatchat/javascript-client-sdk/pull/127 due to the configuration currently not being loaded automatically by the javascript client sdk.

Resolves #759

Tested by @alice_werefox and @dadadah

### DIFF
--- a/packages/client/components/auth/src/flows/FlowCreate.tsx
+++ b/packages/client/components/auth/src/flows/FlowCreate.tsx
@@ -1,7 +1,8 @@
 import { Trans } from "@lingui-solid/solid/macro";
 
-import { useApi, useClient } from "@revolt/client";
+import { useApi, useClient, useClientLifecycle } from "@revolt/client";
 import { CONFIGURATION } from "@revolt/common";
+import { useModals } from "@revolt/modal";
 import { useNavigate, useParams } from "@revolt/routing";
 import { Button, Row, iconSize } from "@revolt/ui";
 
@@ -20,6 +21,8 @@ export default function FlowCreate() {
   const getClient = useClient();
   const navigate = useNavigate();
   const { code } = useParams();
+  const modals = useModals();
+  const { login } = useClientLifecycle();
 
   /**
    * Create an account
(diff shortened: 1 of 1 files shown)


Bob-Swan:
I tested an alternative approach that avoids relying on client SDK config preload timing: after /auth/account/create, fetch GET / and branch on eatures.email directly in FlowCreate.\n\nOpened here for comparison: https://github.com/stoatchat/for-web/pull/853\n\nHappy to close if you'd rather keep everything in this PR and I can adapt it there instead.

packages/client/components/auth/src/flows/FlowCreate.tsx
```diff
@@ -34,13 +37,19 @@ export default function FlowCreate() {
...
+    } else {
+      await login(
+        {
+          email,
+          password,
+        },
+        modals,
+      );
+      navigate("/login/auth", { replace: true });
+    }
```
insertish: I think, in case there is an issue with loading configuration (really this should be rebuilt and made 100% reliable), I'd prefer if we show the email flow if (configuration is not present) OR (email is enabled).

insertish (dismissed):

packages/client/components/auth/src/flows/FlowCreate.tsx
Dadadah: Good catch, I've reversed that statement and added a configuration presence check.

chrishultin (approved):

gabrielfordevelopment (approved):

Issue #741, opened 2026-02-14
Message box doesn't have maximum height set


sirlan-ff00ff:
### What happened?

Using the main app at `https://stoat.chat/` as of 14 Feb 2025

The message box doesn't seem to have a maximum height set, so if your text has a lot of lines, it will grow to take over the whole chat view

<img width="854" height="1056" alt="Image" src="https://github.com/user-attachments/assets/870fb949-31f2-4de0-9b72-e692f18f5fce" />

---

Playing around a bit with CSS, it seems to be a pretty trivial fix by bounding the message box container and adding `max-height: 100%` to one of the children, i.e.

<img width="1489" height="819" alt="Image" src="https://github.com/user-attachments/assets/d5465999-67f2-4086-a354-b590abbcdd8c" />

Pecacheu:
Fixed by #762

PR #762, opened 2026-02-14
fix: give message box a maximum height

sirlan-ff00ff:
Fix for #741


Pecacheu:
This seems like a good idea

sirlan-ff00ff:
so... anything needed to change for this to be merged?

PR #764, opened 2026-02-15
fix: UI improvements for mobile & small screens

Pecacheu:
Made a few UI improvements for Stoat on phones and tablets, namely:
- Replaced obnoxious voice call card with a button in the header bar a la Discord
- Shrink home-screen layout to single-column for mobile
- Settings menus replace the "ESC" button on the right with a back button in the left menu for mid-size screens (eg. tablets)
- Text editor overflows placeholder text with ... as Discord and Telegram do, but expands it once you start typing
- Text editor allows multi-line input with enter on mobile instead of sending
- Fix for button touch event propagation bug causing touches to glitch through modal as it closes

Planning on more mobile UI improvements and a portrait phone layout, but this is what I gots for now.

### DIFF
--- a/packages/client/components/i18n/catalogs/ar/messages.po
+++ b/packages/client/components/i18n/catalogs/ar/messages.po
@@ -400,6 +400,7 @@ msgstr "الصورة الرمزية"
 #: components/auth/src/flows/FlowLogin.tsx
 #: components/auth/src/flows/FlowCreate.tsx
 #: components/auth/src/flows/FlowCheck.tsx
+#: components/app/interface/settings/user/_AccountCard.tsx
 msgid "Back"
 msgstr ""
 
@@ -1500,7 +1501,7 @@ msgstr ""
 msgid "Join the Stoat Lounge"
 msgstr ""
 
-#: components/ui/components/features/voice/callCard/VoiceCallCardPreview.tsx
+#: src/interface/channels/ChannelHeader.tsx
 msgid "Join the voice channel"
 msgstr ""
 
@@ -2537,8 +2538,8 @@ msgid "Start Minimised to Tray"
 msgstr ""
 
 #: components/ui/components/features/voice/callCard/VoiceCallCardPreview.tsx
-msgid "Start the call"
-msgstr ""
+#~ msgid "Start the call"
+#~ msgstr ""
 
 #: components/app/interface/settings/user/Native.tsx
 msgid "Start with Computer"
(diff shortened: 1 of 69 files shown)


packages/client/components/ui/components/design/Button.tsx
```diff
@@ -119,10 +119,18 @@ export function Button(props: Props) {
   );
 
   const { buttonProps } = createButton(rest, () => ref);
+
+  function stopProp(e: TouchEvent) {
+    e.preventDefault();
```
acidtib:
```suggestion
```

`e.preventDefault()` in the touch handlers will prevent the click events from firing on buttons on mobile

packages/client/components/ui/components/design/IconButton.tsx
```diff
@@ -40,10 +40,18 @@ export function IconButton(props: Props) {
   let ref: HTMLButtonElement | undefined;
 
   const { buttonProps } = createButton(rest, () => ref);
+
+  function stopProp(e: TouchEvent) {
+    e.preventDefault();
```
acidtib:
```suggestion
```
same as above

packages/client/components/ui/components/design/Button.tsx
Pecacheu: Hmm interesting, it's not having that effect for me in latest Chrome for Android and is preventing the click-through issue. Might need more testing, could be a Firefox/Safari thing?

acidtib: i tested on firefox using `Responsive Design Mode` and also tested on iPhone safari

Pecacheu: Okay, new discovery after testing in Firefox:

On Firefox, the touch propagation issue does not occur to begin with, and preventDefault() causes the buttons not to work right. On Chrome, however, stopPropagation() alone is not sufficient and preventDefault() is required. Could be a Chrome bug?

Hmm, for now I think I'll drop this code from this PR, I'll create an issue and this can be a separate bug-fix PR since it's getting complicated lol.

insertish:
Just a heads-up with voice card preview, I believe it was intended for server voice channels, so it should be kept in that circumstance, otherwise header button is fine.

Also unrelated .gitmodules change, haven't reviewed too thoroughly apart from that. Just wanted to comment on that change.

Pecacheu:
> Just a heads-up with voice card preview, I believe it was intended for server voice channels, so it should be kept in that circumstance, otherwise header button is fine.
> 
> Also unrelated .gitmodules change, haven't reviewed too thoroughly apart from that. Just wanted to comment on that change.

Yeah, the voice card preview being only for voice channels on servers makes a lot of sense, that should be an easy change.

The gitmodules change is because using `ssh://git@github.com/stoatchat/assets` it fails to pull/initialize the submodule, but it works fine via https, I saw some others mention that too when they were trying to clone the repo.

Pecacheu:
@insertish Just addressed that in my other PR. Easier to keep track of just one since that one integrates the changes from this one anyway.

PR #769, opened 2026-02-15
feat: add invite field to creation flow

kasivulpis:
This PR has been tested on my own instance and is ready for review!

image
image
image
image

### DIFF
--- a/packages/client/components/common/lib/env.ts
+++ b/packages/client/components/common/lib/env.ts
@@ -88,5 +88,5 @@ export default {
   /**
    * Whether or not to show the invite field in the account creation form.
    */
-  INVITE_ONLY: import.meta.env.VITE_INVITE_ONLY as string
+  INVITE_ONLY: import.meta.env.VITE_INVITE_ONLY as string,
 };
--- a/packages/client/components/auth/src/flows/Form.tsx
+++ b/packages/client/components/auth/src/flows/Form.tsx
@@ -9,7 +9,13 @@ import { Checkbox2, Column, Text, TextField } from "@revolt/ui";
 /**
  * Available field types
  */
-export type Field = "email" | "password" | "new-password" | "log-out" | "username" | "invite";
+export type Field =
+  | "email"
+  | "password"
+  | "new-password"
+  | "log-out"
+  | "username"
+  | "invite";
 
 /**
  * Properties to apply to fields


packages/client/components/auth/src/flows/FlowCreate.tsx
```diff
@@ -27,20 +34,21 @@ export default function FlowCreate() {
     const email = data.get("email") as string;
     const password = data.get("password") as string;
     const captcha = data.get("captcha") as string;
+    const body: API.DataCreateAccount = { email, password, captcha };
+
+    if (CONFIGURATION.INVITE_ONLY) {
+      body.invite = data.get("invite") as string;
```
seva-luchianov, line 40: Could I propose looking for an `?invite=code` qs and setting `invite` to that value if `data.get("invite")` returns an empty string? It would make the signup flow for invited users a lot more straightforward.

Alternatively the qs code could prefill the invite field.

Either way this results in the user being registered not having to know anything about invite only functionality and/or copy pasting codes, but also gives them the flexibility to easily override the invite code from the qs with whatever they want if necessary.

Kirskov:
Ah! I had the same problem, and this PR will finally solve it !

packages/client/components/i18n/catalogs/fr/messages.po
```diff
@@ -1216,6 +1216,10 @@ msgstr "Entrez votre mot de passe actuel."
 msgid "Enter your current password..."
 msgstr ""
 
+#: components/auth/src/flows/Form.tsx
+msgid "Enter your invite code."
+msgstr ""
```
Kirskov, line 1221: In french it would be: 

"Entrez votre code d'invitation"

packages/client/components/auth/src/flows/FlowCreate.tsx
Kirskov: Yes, It would be easier if the link was to be copied in the browser then the invite code is passed with the http body.

That way we would not need to copy or type de code in a field. But for now if this solution works I will take it gladly :D

packages/client/components/common/lib/env.ts
```diff
@@ -85,4 +85,8 @@ export default {
   DEVELOPMENT_USER_ID: import.meta.env.DEV
     ? (import.meta.env.VITE_USER_ID as string)
     : undefined,
+  /**
+   * Whether or not to show the invite field in the account creation form.
+   */
+  INVITE_ONLY: import.meta.env.VITE_INVITE_ONLY as string,
```
chrishultin, line 91: Let's update this to pull from the configuration - "invite only" should be exposed now.

packages/client/components/auth/src/flows/FlowCreate.tsx
chrishultin: Agreed - we ought to pull this from the URL if it's available, to make the flow more convenient.

chrishultin (changes requested):

packages/client/components/auth/src/flows/FlowCreate.tsx
Kirskov: @chrishultin I propose this : https://github.com/stoatchat/for-web/pull/957, it works on my self hosted setup.

packages/client/components/common/lib/env.ts
Dadadah: Just a heads up, invite_only isn't reactive, you'll need to use a derived signal that reads from client.configured(). If you need more explanation for that feel free to @ me.

Kirskov: @Dadadah I've implemented a simpler solution where we just need in the backend a script that will generate the invite link in here : https://github.com/stoatchat/for-web/pull/957

it's literraly 2 lines

kasivulpis: > Let's update this to pull from the configuration - "invite only" should be exposed now.

Sorry I'm out of the loop since I've been on vacation. Could you point me to a pr/change so I understand what you mean better @chrishultin?

packages/client/components/auth/src/flows/FlowCreate.tsx
Misofist: In stoatchat/for-legacy-web, the query parameter is called `code`. Should we consider using the same name for compatibility?

Kirskov: @Misofist I can make the change in my PR yes. As long as the backend has a client that create the right url (I made a shell script to do that).

packages/client/components/common/lib/env.ts
Kirskov: > @Dadadah I've implemented a simpler solution where we just need in the backend a script that will generate the invite link in here : #957
> 
> it's literraly 2 lines

@Dadadah Can you please elaborate your thumb down for my proposal?

Dadadah: > > Let's update this to pull from the configuration - "invite only" should be exposed now.
> 
> Sorry I'm out of the loop since I've been on vacation. Could you point me to a pr/change so I understand what you mean better @chrishultin?

Please merge main into your branch. The stoat.js dependency has been updated. You can now check if invite only is enabled by using the client, instead of adding it to env.ts. For an example of how to do it, you can see [here](https://github.com/Dadadah/stoat-for-web/blob/feature/invite_only/packages/client/components/auth/src/flows/FlowCreate.tsx).

kasivulpis:
As I'm sure people can see I've been too busy to contribute to Stoat and update my PR (considering that my last response was a week ago.) I'm closing this PR in favor of @Dadadah's PR since I keep being too busy to fix it.

PR #775, opened 2026-02-15
Add Dutch translations

Iemand005:


reeesespuffs:
Translations should not be added manually, translations are done through [here](https://translate.stoat.chat/projects/revolt/revolt-for-web/)

Issue #781, opened 2026-02-15
Clicking GIF menu too many times fastly causes button to become unclickable


guizaa:
### What happened?

It seems that the https://api.gifbox.me/categories?locale=en_US url that is loaded by the GIF button can send a 429 error response which causes the GIF button to be unclickable until you refresh the page.

Dadadah:
This should have been fixed a few months ago, if you run into this again please open another issue.

PR #782, opened 2026-02-15
fix: add response code check to GifPicker

guizaa:
To avoid HTTP errors causing the GifPicker to not load when clicked, retry requests on 429 errors and return empty object when unrecoverable errors occur.

Fix for #781

### DIFF
--- a/packages/client/components/app/interface/settings/user/bots/ViewBot.tsx
+++ b/packages/client/components/app/interface/settings/user/bots/ViewBot.tsx
@@ -6,6 +6,7 @@ import { useModals } from "@revolt/modal";
 import { CategoryButton, Column, iconSize } from "@revolt/ui";
 
 import MdContentCopy from "@material-design-icons/svg/outlined/content_copy.svg?component-solid";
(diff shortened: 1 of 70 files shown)


guizaa:
@reeesespuffs The commits you added seem to break the original fix. Is there a reason the named lambda `getTrendingCategories()` was removed?

packages/client/components/ui/components/features/messaging/composition/picker/GifPicker.tsx
```diff
@@ -106,7 +106,19 @@ function Categories() {
...
+          .then(async (r) => {
+            const data = await r.json();
+            if (r.ok) return data;
+            else if (r.status == 429) {
+              setTimeout(getTrendingCategories, data.retry_after);
+            }
+            else return {};
+          })
+          .catch((_) => {return {}});
+      }
```
insertish, line 119: I don't super like this solution, it feels a bit too bulky for what should really be abstracted & ideally we actually start doing better 429 handling across the app generally.

Maybe there's some way to just hook the retry_after into https://tanstack.com/query/latest/docs/framework/solid/guides/query-retries? (and maybe make this a common pattern across the app?)

PR #783, opened 2026-02-16
feat: Add Toggleable Enhanced Noise Suppression Using RNNoise

Dadadah:
### Summary
This PR adds enhanced noise suppression using RNNoise. It adds a package that I have refactored and put up on npm. Here's the link to the package: 
https://www.npmjs.com/package/livekit-rnnoise-processor

The RNNoise model + worklet is a little under 2MB, it only downloads when a user connects to voice with Enhanced Noise Suppression enabled.

### What is RNNoise and how does this PR use it
RNNoise is a noise suppression library based on a recurrent neural network. It is open source under the BSD-3 license and the source can be found [here](https://gitlab.xiph.org/xiph/rnnoise). This PR uses RNNoise by downloading via CDN a WASM compilled lite model of the library. The library is compiled with Emscripten, added to a worklet, and hosted by CDN. The worklet is then sent to the [AudioWorklet API](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet) whenever the processor is used in a livekit track. The worklet takes packets from the input, resamples the samplerate to 48khz as required by RNNoise, sends the packets through RNNoise, then resamples to it's original format and sends it back to the worklet api.

### Notes
~~This package requires importing the following RNNoise model from cdn: [dist/DenoiserWorklet.js](https://cdn.jsdelivr.net/gh/dadadah/livekit-rnnoise-processor@afa25e08a9a57cfc859175e42c71c7b46c2e41a1/dist/DenoiserWorklet.js). This points to my github repo using jsdelivr. The package allows passing a different CDN URL on processor creation. The worklet can be compiled by a maintainer, uploaded to a Stoat CDN, and passed to the processor so it will load that version instead.~~ This has been addressed by allowing configuring the CDN location via vite.

Adds a noise suppression select.

One of the changes here is a change of when the microphone channel is enabled. Previously, the microphone channel was enabled immediately upon connecting to livekit. This caused the audioContext on the microphone stream to be overwritten with an undefined context. Processors require an audio context. In this pr, I have made it wait until the client is connected before enabling the microphone stream. This makes a muted icon appear on the user icon when joining a voice channel until they are fully connected.

This package introduces a memory leak on chromium based browsers. It will leak about 400kb per channel join. More details about the chromium bug can be found [here](https://issues.chromium.org/issues/40072701). I have limited the memory leak as much as I can, but it will not be possible to do this without the memory leak on chromium based browsers without a serious refactor to how Stoat handles microphone inputs.


Dadadah:
The livekit-rnnoise-processor package should not have webpack as a dependency, I'll fix that this evening.

Dadadah:
Removed webpack dependency, removed the translations as I didn't know that process was automatic (new to a lot of this tooling), added the option to pass in a CDN URL to the processor to allow a safe worklet host via Stoat's maintainer team.

Dadadah:
Here's a few pictures to showcase the new setting.
(pretend the typo for RNNnoise isn't there, I fixed that)

image
image
image

Dadadah:
I have found that some users are unable to use the enhanced noise suppression. I'm going to do some debugging but it may take me some time to get to the bottom of it. I've marked this PR as a draft in the mean time as I don't want to merge it with this bug. If a week goes by and I still can't figure it out, we can consider merging it as is and add a disclaimer that it won't work for all users as this is a desired feature.

Dadadah:
Bug has been resolved by implementing a basic mono resampler.

chrishultin (approved):

mihaicm93 (approved):

PR #791, opened 2026-02-16
chore: add human-readable css class names to components

amycatgirl:
this changes a bunch of files and is based from an older commit. going to rebase it in a moment.

stacked on top of #646


Pecacheu:
Sorry for the commit spam, it's caused by rebasing commits mentioning this lol

But yeah some of my PRs required defining classes on elements, so I'm trying to maintain the naming conventions used here so that if the CSS feature ever drops eventually everything will have sensible class names and it shall be glorious

PR #801, opened 2026-02-17
fix: Limit the length of various fields

MHanak1:
Addresses part of #710, supersedes #726

Limits the length of fields for:
* Channel name
* Channel description
* Server name
* Server description
* Emoji name
* Role name
* Ban reason
* Username
* Bot username
* Display name
* Nickname
* Category name
* Group name
* Webhook name
* Status

This commit also adds the required props from `mdui` to the `Form2.TextField` component


MHanak1:
I should add that my use of the `counter` prop is fairly inconsistent

Edit: now all fields show the character counter

reeesespuffs:
Heads up! DCO is failing, please remember to sign off your commits

MHanak1:
> Heads up! DCO is failing, please remember to sign off your commits

Sorry about that, i have fixed both that, and the first commit not following conventional commit style. I really do need to get used to doing that

MHanak1:
This might conflict with #723 which also sets a character limit on the emote field in a slightly different way. I'm not sure whether it's better to set the limit through the `maxlength` attribute or the `maxLength` property (or do they end up doing the same?), but i think using `counter` might be a better solution

mihaicm93:
> This might conflict with #723 which also sets a character limit on the emote field in a slightly different way. I'm not sure whether it's better to set the limit through the `maxlength` attribute or the `maxLength` property (or do they end up doing the same?), but i think using `counter` might be a better solution

I think with `counter` is a better way to handle it, about the `maxlength` and `maxLength` I belive they do same thing but I would be interested to know as well wich is prefered

mihaicm93 (approved):

gabrielfordevelopment:
Would be nice to use constants instead of magic numbers.
Like, for most of these you use min-lenght 1 or 2 and max-lenght 32.

But if later we change the length then we have to manually re-write all the files, and we might miss some of them, leading to problems and inconsistency.

So it would be beneficial for maintaining the code to have constants for min and max lenght and use those.

MHanak1:
> Would be nice to use constants instead of magic numbers.
> Like, for most of these you use min-lenght 1 or 2 and max-lenght 32.
> 
> But if later we change the length then we have to manually re-write all the files, and we might miss some of them, leading to problems and inconsistency.
> 
> So it would be beneficial for maintaining the code to have constants for min and max lenght and use those.

I'm not sure if i agree, this is similar to how this is done in the backend (with the `#[validate]` macros), and having the frontent use constaints would make it further from how it is implemented in the backend

PR #803, opened 2026-02-17
feat: Redirect to '/app' after successful deletion of the server

0xAlcidius:
When a user deletes a server. The user stays within the server's settings view. This PR will redirect the user to '/app'.


mihaicm93:
While this does redirect users to /app in browser as seen below 
firefox_Yw3EPt9vhc

On desktop client it does not work 
electron_ibNmJl0Awa

seems like the issue is a bit more difficult to fix then just a reddirect from my digging into it

mihaicm93:
https://github.com/stoatchat/javascript-client-sdk/pull/134 Fixes this issue at it's core

PR #835, opened 2026-02-19
feat: Mobile UI for web

### LINKED ISSUE #850: Impossible to leave a server on mobile
### What happened?

On the mobile web app its currently impossible to leave a server as theres no button for a drop down nor ability to right click. Holding the icon just opens the server.

### LINKED ISSUE #883: Body text overflows past edge of error dialog and other modals
### What happened?

Surprised this doesn't already have an issue reported, but this can happen in particular with error dialogs since they often contain long strings of text without spaces to break it up.

I'm about to fix this in my PR but I wanted to report the bug first just so it's there.

### LINKED ISSUE #887: UI not rescaling in PWA for Android
### What happened?

Environment: Pixel 8a - grapheneOS
Client: PWA via vanadium browser
Steps to reproduce: Log into PWA and load a server page.

Result:
The UI panes/windows for server, server chats, and server user list, all crowd each other. They force each of them to shrink and not leave enough space to view their contents.

Expectation:
Expected the UI panes to hide, or scale, so that they could be viewed and used as would be needed to read chats, join calls, or do other normal navigation around the app.

This is my first time submitting an issue, if I submitted incorrectly please let me know the preferred method or guidelines for issue submission, and I'll do my best to accommodate.

### LINKED ISSUE #924: web app ui scaling broken on android
### What happened?

Image
Image

### LINKED ISSUE #998: Skip this prompt

### LINKED ISSUE #1005: login/sign-up/reset flows: default cursor displayed when hovering over editable text input fields
### What happened?

When hovering over the text input fields in the login/sign-up/password reset flows, the default cursor is displayed instead of the text cursor.

Version: observed in 0.2.1 (not tested in earlier versions) and reproduced in 0.3.0

## Actual behavior

Image

Image

Image

## Expected

A text cursor is displayed, as is the case when hovering over an editable message text field

Image

### LINKED ISSUE #1023: Editing a message crops the auto-complete
### What happened?

Editing message → autocomplete gets cropped down

Image

---

UI/UX issue - good luck for the person trying to get his PR approved.

### LINKED ISSUE #1033: font selector dropdown menu glitched on desktop app and browser
### What happened?

when trying to scroll the dropdown menu to select the font it glitched to the side and started blinking, making it unusable
Browser version did not have this issue back when i made [the original post in the wrong place](https://github.com/stoatchat/for-web/issues/661), however I am now able to trigger the same bug on browser too. I do not have the permissions to update the other ticket though.

demonstration:

https://github.com/user-attachments/assets/a3bd2d47-ff65-4573-9e07-31e55ac1456a

Edit: the video here was shot at a higher framerate but brought down to 30fps during the file compression, thus not showing the fast blinking of the dropdown menu that adds to the abnormal list position.

### LINKED ISSUE #1207: Mobile app incorrectly loads squished 3-pane desktop layout on phone screen
### What happened?

Upon opening the Stoat mobile app, the user interface does not load the standard single-pane mobile layout. Instead, it forces the 3-pane desktop/tablet layout (showing the Server list, Channel list, and Chat box all at the same time).Because the mobile screen is narrow, all three columns are severely squished side-by-side. This makes each section appear half-cut, unreadable, and impossible to navigate.
Expected Behavior:The app should automatically detect a mobile screen size and display a single-pane layout where sidebars are hidden and accessible via swiping or a menu button. 
(I am here to help this community more bigger. Therefore I would like to give you each and every info about the app and try my level best to make it number one.

Image

Pecacheu:
# Prerequisites
- #1025
- #1102
- #1196
- #1092
- #1273
- #1216
- #1253

# Why this is important
One of the **most consistent** reasons I hear stopping/making folks hesitate from joining Stoat is the borked mobile app situation. A lot of people in my generation spend much of our time on our phones, plus when it comes to keeping in touch at work, it's all you have access to.

Unfortunately, the current state of affairs is that the Android beta app is missing a ton of features and doesn't work right, and the iOS one isn't currently available. The best thing we've been telling these people is "use the web app in your browser instead," but the trouble is, well you get an experience like this (these are screenshots from real users in the Stoat server who are asking for this, as well as one of my friends testing this fork.)

# Responsive mobile UI layout

### Phone Example (S25 Ultra)
The most important UI change is the slide gesture to open and close panels instead of showing them side-by-side. I've done some work to refine this to make it feel as natural as possible, eg. taking into account inertia of the finger to "snap" the panel open and shut.

https://github.com/user-attachments/assets/f7887f7b-ba59-480b-8b74-454200bf17eb

I've un-rounded the corners to increase usable space under a certain display size (opinions welcome on this). I also added code to make the panel slide over automatically when tapping a DM/channel. Lastly, the "channel collapse" button is replaced with a "back" button that returns to the channels view.

### Tablet Example (iPad Mini)
When the screen is big enough, however, the UI dynamically changes back to the traditional layout. However, the "server members" panel on the right is collapsed by default in this case, as we found it was creating some confusion among testers. One user also suggests allowing a swipe gesture to open that on tablets, but open to thoughts on that.

# Other enhancements for touch users

- Enable autocapitalize and autocorrect on message box (these can be configured as desired in the user's OS keyboard settings, but not enabling them here prevents them from working regardless of that setting.)
- Change keyboard shortcuts so that `Ctrl + Enter` sends on mobile instead of `Enter`, enabling newlines on touch-keyboard without sending.
- Disable right-click/long-press context menu when editing a message, so that spell-check suggestions can actually be clicked on.
- Don't trigger floating tooltops or message toolbar on touchscreens, otherwise they get stuck due to no mouse events.
    - Added "React" button to context menu to still allow reacts on mobile.
- Skip User Card preview when tapping user icon on mobile and go straight to full User Profile.
- Adjust margins and min sizes on modals, invite previews, and some other UI elements to improve the mobile experience.

## Linked Issues

Fixes #850
Fixes #883
Fixes #887
Fixes #924
Fixes #998
Fixes #1005
Fixes #1023
Fixes #1033
Fixes #1207

# TODO
- [x] Remember last SlideDrawer state (open/closed) when rotating to landscape then back to portrait. Don't necessarily need to persist this in the data store.
- [x] Auto-hide touch keyboard when swiping from messages view to channels view.
- [x] Search icon instead of input box in header on mobile (phone & tablet layouts).
- [x] Make channel name ellipsis truncated to ensure header buttons fit.
- [x] When dragging finger over top of a channel/DM and sliding the drawer, the channel/DM button can get visually stuck in the 'hovered' state.
- [x] Transition layout code from using native CSS media queries to PandaCSS breakpoints where possible, and reactive vars from `useDevice` otherwise.
- [ ] Hold-and-drag to reorder server sidebar and channel bar doesn't work good on touchscreens.
- [ ] Settings menu will be moving to a full-screen modal, which may necessitate code changes.
    - Likely still using SlideDrawer for navigation in settings?
- [ ] Use a full-screen modal for search and members panel on mobile instead of the sidebar.
- [ ] Appearance menu needs... something, idk. Maybe use dropdowns instead in mobile layout?

Test live at https://forestfire.net:420

_Btw, quick thanks to everyone who's helped test this so far on lots of different devices and helped fix lots of bugs!_


Pecacheu:
Okay I'm going to attempt to fix the DCO check and i18n again while pushing some commits. If I don't come back, play Porter Robinson at my funeral.

Pecacheu:
TODO Working on splitting off any unrelated/semi-related features or out-of-scope bugfixes to their own PRs, may take a bit :)

Misofist:
I think this topbar within a channel needs some improvement, especially with the search bar. The biggest issue is that the channel name can push things off the right hand size of the screen.

{F48A47DA-330D-4D95-8E9E-4DE63DD81A60}
{308460D2-8736-4A74-A41D-927F8DE9AFF9}
{CB2FB10D-2268-413C-9850-A7512CC787B5}

I think that the buttons in the top right should be fixed to the side of the screen, and the channel name should be ellipsis-truncated if it does not fit beside them.

I also think that the search bar with placeholder text could be replaced with a iconographic search button, and tapping this button could then cause the search bar to expand to the left, covering the other buttons.

I also think that when the search results, members list, or pinned messages are displayed they should take up the entire screen, and should have a back arrow in the top left.

Other than that, this UI seems very useable. I'm using it on my instance and haven't noticed any major issues.

infi (approved):
I don't see our mobile strategy blocking this atp so LGTM in terms of UI/UX

Pecacheu:
> I don't see our mobile strategy blocking this atp so LGTM in terms of UI/UX

Epic. Few things on the technical side to iron out before merge though.

@Dadadah suggested we centralize reactive hooks for layout changes, which will be dealt with in #1196

Also a few screens do be looking a little rough at certain screen sizes- Eg. the Appearance menu has been brought up to me as less than ideal. (*This is at tablet size riiiight before it hides the left sidebar)

Dadadah (changes requested):
I'm not comfortable merging this PR. I've spoken with @Pecacheu a lot about it, and I've been working with them to break out components from this PR to make it more manageable to review. This PR is 2000 lines, and I am sure there are many aspects of this PR we can break out into separate PRs. Don't get me wrong, I want these mobile changes available in for-web, but I think one monolithic PR is not the move.

I'm leaving this review here so this doesn't get merged.

Edit: With @infi's approval, I will fast-track any changes that are pulled from this PR past UI/UX review to get features merged faster.

packages/client/components/app/interface/settings/_layout/Content.tsx
```diff
@@ -19,34 +19,42 @@ export function SettingsContent(props: {
...
+    contRef!.classList.add(...scrollBase.split(" "));
+  });
 
   return (
     <div
-      use:scrollable={{
-        class: base(),
-      }}
+      ref={(ref) => props.ref((contRef = ref))}
+      use:scrollable={{ class: base }}
```
Dadadah: This seems like a reactivity issue. In solid, you should be calling the class as a function if it's going to change. Why are you calling this on mount?

packages/client/components/ui/components/features/messaging/composition/picker/EmojiPicker.tsx
Dadadah, line 1: I know it's not great, but can you remove the refactor to this file as to not cause issues with the other emojipicker refactor that's coming in?

packages/client/components/ui/components/features/messaging/composition/picker/GifPicker.tsx
Dadadah, line 1: I know it's not great, but can you remove the refactor to this file as to not cause issues with the other gif picker refactor that's coming in?

packages/client/src/interface/channels/ChannelHeader.tsx
```diff
@@ -251,27 +256,48 @@ export function ChannelHeader(props: Props) {
...
+                },
+              }}
+            >
+              <Symbol>search</Symbol>
+            </IconButton>
           }
-        />
+        >
+          <input
+            class={css({
```
Dadadah: Please extract this css call into another variable.

packages/client/vite.config.ts
```diff
@@ -33,13 +33,16 @@ export default defineConfig({
       injectManifest: {
         maximumFileSizeToCacheInBytes: 4000000,
       },
+      devOptions: {
```
Dadadah: What does this do? Should it be turned off in production?

Dadadah (changes requested):

packages/client/components/app/interface/settings/_layout/Content.tsx
Pecacheu: Sorry for the bizarre solution here, not sure if there's a better way but this is actually *to* make it reactive. With the standard method I was getting weird issues due to use:scrollable (I added a comment mentioning that's why). It seems like the code in use:scrollable makes an attempt to *not* overwrite the previous classes, but for whatever reason, it does anyways. Maybe this is also the cause behind use:scrollable not playing nicely with styled() elements.

packages/client/src/interface/channels/ChannelHeader.tsx
Pecacheu: This code was already there (I just changed the indentation), but yeah I think that or a styled element makes sense instead of using css() inline

packages/client/vite.config.ts
Pecacheu: This is a weirdly named Vite config option that enables the PWA in dev mode, it has no effect when run in production mode where the PWA is always enabled.

It was necessary for not loosing my mind while developing the mobileui PR by getting the PWA to work in dev mode instead of having to rebuild the whole app every time I made a change that needed testing in PWA mode, but it really should be in a separate little `chore:` PR or something.

packages/client/components/app/interface/settings/_layout/Content.tsx
Dadadah: can't you use `<div class={base} ...` instead? It looks like the friends list uses that and it works as expected.

Pecacheu: Hmm... Not exactly, it looks like FriendsList had the same problem with use:scrollable as they used another unusual solution here. They gave it a class name:
`<div class="FriendsList" ref={scrollTargetElement} use:scrollable>`

And then in a parent element they define the style of it *indirectly*, instead of directly using a styled() element or css() style:
```js
"& .FriendsList": {
  height: "100%",
  paddingInline: "var(--gap-lg)",
},
```

I'd try that approach, but in this case the element is at the highest level in its respective solid JS function.

Dadadah (approved):

PR #848, opened 2026-02-20
feat: User card redesign

MHanak1:
This PR redesigns the profile card so it is more compact and gives more space for relevant items, making the bio readable without having to scroll

It would be great if I could get some feedback on the design:

# Old
Screenshot 2026-02-20 at 10-24-35 Stoat
Screenshot 2026-02-20 at 19-32-18 Stoat

# New
image
image

### DIFF
--- a/packages/client/components/modal/modals/UserProfile.tsx
+++ b/packages/client/components/modal/modals/UserProfile.tsx
@@ -27,11 +27,8 @@ export function UserProfileModal(
           bannerUrl={query.data?.animatedBannerURL}
         />
 
-        <Profile.Status user={props.user} />
-        <BadgeAndActionsRow>
-          <Profile.Badges user={props.user} />
-          <Profile.Actions user={props.user} />
-        </BadgeAndActionsRow>
+        <Profile.Badges user={props.user} />
+        <Profile.Actions user={props.user} />
         <Profile.Bio content={query.data?.content} />
         <Profile.Joined user={props.user} />
       </ProfileContents>
@@ -45,14 +42,5 @@ const ProfileContents = styled("div", {
     flexDirection: "column",
     gap: "var(--gap-md)",
     padding: "var(--gap-md)",
-  },
-});
-
-const BadgeAndActionsRow = styled("div", {
-  base: {
-    display: "flex",
-    gap: "var(--gap-sm)",
-    justifyContent: "space-between",
-    alignItems: "center",
   },
 });
--- a/packages/client/components/ui/components/features/profiles/ProfileActions.tsx
+++ b/packages/client/components/ui/components/features/profiles/ProfileActions.tsx
@@ -44,26 +44,32 @@ export function ProfileActions(props: { user: User; member?: ServerMember }) {
 
   return (
     <Actions>
-      <Show when={props.user.relationship === "None" && !props.user.bot}>
-        <Button onPress={() => props.user.addFriend()}>Add Friend</Button>
-      </Show>
-      <Show when={props.user.relationship === "Incoming"}>
-        <Button onPress={() => props.user.addFriend()}>
-          Accept friend request
-        </Button>
-        <IconButton onPress={() => props.user.removeFriend()}>
-          <MdCancel />
-        </IconButton>
-      </Show>
-      <Show when={props.user.relationship === "Outgoing"}>
-        <Button onPress={() => props.user.removeFriend()}>
-          Cancel friend request
-        </Button>
-      </Show>
-      <Show when={props.user.relationship === "Friend"}>
-        <Button onPress={openDm}>Message</Button>
-      </Show>
-
+      <div style={{ "margin-right": "auto" }}>
+        {" "}
+        <Show when={props.user.relationship === "None" && !props.user.bot}>
+          <Button size="sm" onPress={() => props.user.addFriend()}>
+            Add Friend
+          </Button>
+        </Show>
+        <Show when={props.user.relationship === "Incoming"}>
+          <Button size="sm" onPress={() => props.user.addFriend()}>
+            Accept friend request
+          </Button>
+          <IconButton size="sm" onPress={() => props.user.removeFriend()}>
+            <MdCancel />
+          </IconButton>
+        </Show>
+        <Show when={props.user.relationship === "Outgoing"}>
+          <Button size="sm" onPress={() => props.user.removeFriend()}>
+            Cancel friend request
+          </Button>
+        </Show>
+        <Show when={props.user.relationship === "Friend"}>
+          <Button size="sm" onPress={openDm}>
+            Message
+          </Button>
+        </Show>
+      </div>
       <Show
         when={
           props.member
@@ -76,12 +82,13 @@ export function ProfileActions(props: { user: User; member?: ServerMember }) {
             : props.user.self
         }
       >
-        <IconButton onPress={openEdit}>
+        <IconButton size="sm" onPress={openEdit}>
           <MdEdit {...iconSize(16)} />
         </IconButton>
       </Show>
 
       <IconButton
+        size="sm"
         use:floating={{
           contextMenu: () => (
             <UserContextMenu user={props.user} member={props.member} />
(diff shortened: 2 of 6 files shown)


Pecacheu:
I'm a fan of the new layout ngl
Simpler and easier to visually follow but still on-theme

World170:
Loving the progress so far!

RatDotRaw:
To add a suggestion to how roles are in the new design. They overflow, and I'm assuming there will be scrolling behavior.

I suggest just wrapping roles. As out of experience, without any clear indication, people will not realize this. 
This can lead to a sort of obfuscation of roles.

And if you decide to keep the overflow, add little scroll arrows to click on (or at least a scroll bar).

Just something to consider.

---

As for a more personal suggestion/standpoint. 

I don't like the idea of covering the banner. As new design image 2 demonstrates;
It clutters and hides personalization, which is important to me.

From a more practical standpoint, banners are unpredictable, someone might put a white banner, and all of a sudden your white text is invisible or in general other contrast problems. This is a design angle I have struggled before with.
And out of my experience, the easiest solution is to just not to put text over unpredictable images.

gabrielfordevelopment:
> To add a suggestion to how roles are in the new design. They overflow, and I'm assuming there will be scrolling behavior.
> 
> I suggest just wrapping roles. As out of experience, without any clear indication, people will not realize this. This can lead to a sort of obfuscation of roles.
> 
> And if you decide to keep the overflow, add little scroll arrows to click on (or at least a scroll bar).
> 
> Just something to consider.
> 
> As for a more personal suggestion/standpoint.
> 
> I don't like the idea of covering the banner. As new design image 2 demonstrates; It clutters and hides personalization, which is important to me.
> 
> From a more practical standpoint, banners are unpredictable, someone might put a white banner, and all of a sudden your white text is invisible or in general other contrast problems. This is a design angle I have struggled before with. And out of my experience, the easiest solution is to just not to put text over unpredictable images.

Agree. Putting text on the banner doesn't feel good. Some people tend to put banner that already have some kind of text which makes it even more unreadable.
Also the stuff with the roles... I don't want to horizontally scroll to see all of them.
It still requires adjustments imo.

uaxdotnet:
I was previously trying to make it look like this

image

Although I abandoned it since there is no central design document and I didn't want to refactor multiple data objects which was necessary here.

See if this helps a bit.

MHanak1:
Once i have the time i'll implement the suggestions @RatDotRaw gave. i did consider having the roles wrap before, i am not exactly sure why i didn't try that, as it is really not that obvious that those horizontally scroll, and people who do not use gestures can't even do it. When it comes to the banner, I haven't thought about the aspect of readability, so i might handle the status similarly to how @uaxdotnet did

gaberashko:
I think the suggestion by @uaxdotnet to separate the status from the banner is best. I definitely think the vertical flex structure is the most pleasing to navigate. 

If possible, I would also consider making the profile content learner by replacing explicit names in the "Joined" category (e.g. "Stoat", "dvlkdsk") with icons (Stoat's logo to replace "Stoat", and then the server's icon to replace the "dvlkdsk"), and making them inline to further clean up vertical-bloat in the UserCard component. Instead of having 4 vertical lines of text, you could easily shrink it down to 1 line with such a design.

Otherwise, I think the current progress looks great! Hoping to see it implemented soon!

MHanak1:
So I'm experimenting with how to implement role wrapping and I'm split between two versions: 

Normal Wrap:
image

All but the last line fill the entire width:
image

gabrielfordevelopment:
> So I'm experimenting with how to implement role wrapping and I'm split between two versions:


I like the first version. 
But please don't put anything in the banner side. Remove those badges.
Or try something like this, to only take up the smallest place possible from the banner (Badges right under username, so it is a total of 2 rows of the banner, keeping it minimal):


image


And then if there are too many badges and no place for them, just put a "..." there to indicate that it is not the total amount of badges, or something

MHanak1:
> > So I'm experimenting with how to implement role wrapping and I'm split between two versions:
> 
> I like the first version. But please don't put anything in the banner side. Remove those badges. Or try something like this, to only take up the smallest place possible from the banner (Badges right under username, so it is a total of 2 rows of the banner, keeping it minimal):
> image
> 
> And then if there are too many badges and no place for them, just put a "..." there to indicate that it is not the total amount of badges, or something

That's sadly not possible, as this is where the username is if it is different from the display name. At some point, i did experiment with having the badges show up from the bottom upon hover, but i struggled to implement it well and it wouldn't be very mobile friendly

I don't think i can come up with anything better than just putting them under the banner like so:
image
Or a slightly different version (which i am not sure if i like):
image

RatDotRaw:
Props for listening to feedback! :heart: 

I still have some more suggestion / ideas you can work with reasons behind them.
All in not a specific order, a brain dump if you will.

The little user status icon  (online, focus, ...), I would keep the little dot on the profile picture. People associate it easier, especially if they're coming from discord. They should understand it by cross-referencing other parts anyway.
You can add a hover hint for if users are still in doubt.

Regarding how roles are displayed. Looking at discord, some people have way too many roles (found someone with 18 roles). 
Displaying them all at once is just too much, so some roles are going to need to be limited. I would add a little button to expand / collapse a full view, with a hover hint (with "View all roles").
In the mockup it's the little ball with the +12 in it.

I would make the PFP and banner both a little bigger, letting the users' profile shine :sparkles: 
This is their little space y'know.

Now onto the general order of elements within the card. I would suggest the following order:
1. PFP & banner
2. Custom status (keeping it close to the users' identity)
3. Badges (and maybe pronounce later on)
4. Bio (when users click on profiles, it's mostly to check them out, not to check roles)
5. Roles (Ordered by most important role)
6. Other less important info (like join date)

Where the buttons like `add friend` and such should come, I'm not entirely sure.
Currently, it takes a lot of space, if there are (vertically) smaller buttons available, try those out. Or perhaps use iconography.

The current cart you immediately need to scroll, it's rather annoying while this is not really necessary.
I would make the card vertically longer, make use of the space available.

And ofc I also had some fun making a mockup :>
I tried to make the custom status bubble a bit more playful.
board

Also, is this card a different element form the full-screen one (when you click on the banner)?

uaxdotnet:
For the roles, it's better to limit it to 2 rows like Discord and add a + button which is clickable.

Here is what I've been working on( ignore the badges section but I'll eventually try to incorporate some of the changes here to this PR)

{8A723E2F-C9F0-439B-9692-E97C6BB69B92}

It's still not perfect but the roles management is better since it won't display more than 2 rows.

I just use the existing openRoles function to display the roles but it's handled a bit differently on discord where it just display's all roles at once like in @MHanak1's screenshots.

MHanak1:
@uaxdotnet you can fix the badges by cloning the [assets](https://github.com/stoatchat/assets) repository into `packages/client/assets`

When it comes to the roles, i'll try implementing it myself, and if i fail i'll use your implementation

uaxdotnet:
I've made this act like Discord's Implementation.

You can refer to the code at https://github.com/uaxdotnet/stoat-for-web/tree/user-card-changes

Relevant files are : 

- [packages/client/components/ui/components/features/profiles/ProfileRoles.tsx](https://github.com/uaxdotnet/stoat-for-web/blob/user-card-changes/packages/client/components/ui/components/features/profiles/ProfileRoles.tsx)

Demo : 

{A8D07E0B-40FD-4B64-A9F9-E2030A5C7412}

If you click the '+2' button, it expands the list to show all roles.

{CDDEA8FD-A5B6-49A8-828F-99F98D62F625}

This can be collapsed again by pressing the '<' button.

gabrielfordevelopment:
Thank you so much for taking the PR feedback into account and genuinely considering the suggestions! I really appreciate how thoughtfully you're handling the integration; it's great to see. On behalf of the whole community, thanks for your efforts.

If I may suggest one more thing: I’d love to see the profile picture scaled up a bit. Right now it’s quite small, and unless I’m using a magnifying glass, it’s hard to make out any detail. If you could make it larger - even if it covers a bit more of the banner - I'd be very grateful. Thanks again for all your hard work!

image

infi:
Hi! We are happy to redesign the user card and we definitely see potential in making it better, but at the same time we want to make sure everything in our apps fully and cohesively fits both the vision we are going for as well as the Material 3 design language while avoiding the pitfalls of "design by committee". Therefore we would prefer to internally coordinate a redesign in the future rather than merging this PR. However, rest assured that we will try to factor in all feedback from this thread into such a redesign.

insertish:
Closing out per above.

Issue #824, opened 2026-02-18
Muted server still shows unread message indicator


Rhiven:
### What happened?

When a server is muted it still shows the small unread message indicator to the left of the server icon. If I right click the server and mark all as read it clears it but it comes back again as soon as a message is posted in that server.

OxymoreProducer:
not only some muted channels reappear, but the muted status doesn't seem to transfer across devices. A channel that I muted in the desktop app isn't muted in the browser version nor the android version e.g.

gabrielfordevelopment:
Still a thing. Also having it on `Notification: None` doesn't help.

<img width="201" height="335" alt="Image" src="https://github.com/user-attachments/assets/a571f9c7-0938-462e-b465-bfa840c62da2" />

PR #849, opened 2026-02-20
fix: hide alert if server is muted

meijerivoi:
Fixes bug #824 

First PR, not sure if style is OK

### DIFF
--- a/packages/client/src/interface/navigation/servers/ServerList.tsx
+++ b/packages/client/src/interface/navigation/servers/ServerList.tsx
@@ -258,7 +258,8 @@ export const ServerList = (props: Props) => {
                   indicator:
                     props.selectedServer() === entry.item.id
                       ? "selected"
-                      : entry.item.unread && !state.notifications.isMuted(entry.item)
+                      : entry.item.unread &&
+                          !state.notifications.isMuted(entry.item)
                         ? "alert"
                         : undefined,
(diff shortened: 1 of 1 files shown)


packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -258,7 +258,7 @@ export const ServerList = (props: Props) => {
                   indicator:
                     props.selectedServer() === entry.item.id
                       ? "selected"
-                      : entry.item.unread
+                      : entry.item.unread & !state.notifications.isMuted(entry.item)
```
mihaicm93: I've tested this and works but one small change needed: please change the single `&` to `&&`

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -258,7 +258,7 @@ export const ServerList = (props: Props) => {
                   indicator:
                     props.selectedServer() === entry.item.id
                       ? "selected"
-                      : entry.item.unread
+                      : entry.item.unread && !state.notifications.isMuted(entry.item)
```
mihaicm93: seems like Prettier is not happy with how long this line is and fails, can you run `pnpm prettier --write packages/client/src/**`

PR #861, opened 2026-02-21
fix: role hierarchy enforcement

mihaicm93:
Makes the role management UI actually respect the role hierarchy. Now, you can only drag and reorder roles that are below your own highest role, owner can still modify all the roles

Required -> https://github.com/stoatchat/javascript-client-sdk/pull/128

firefox_EfYNKOJGAB

electron_xxQJnqkaXt


packages/client/components/app/interface/settings/user/Native.tsx
Pecacheu, line 12: I think you may have accidentally left a change from another PR branch in because it doesn't seem(?) like the changes in Native.tsx is related.

mihaicm93: There is a chance I made the branch for this fix right after https://github.com/stoatchat/for-web/pull/802 and probably made the branch from that PRs branch instead of main

Issue #774, opened 2026-02-15
Cannot edit roles on self, even with "manage roles" permission enabled on the highest role in the server.


MidoriMushrooms:
### What happened?

I have the highest role permissions possible to obtain on a Stoat server, but I am not the owner. I cannot assign myself roles in the server at all, even though every other role on the server is below my own. I can edit other users' roles who have a lower role in the hierarchy than myself, but I cannot edit the roles for people who have the same hierarchy as myself, even to remove or assign roles lower than ours in the role order.

I am the admin for a tabletop RPG community and the staff need to frequently assign and clear roles from other people because we create and clear game rooms for different groups all the time. Being unable to set roles onto ourselves means we cannot assign ourselves as participants in games when applicable, and as I am also one of the most prolific GMs, I also need to be able to assign and remove roles from myself often.

It would make more sense for the roles to behave as they do in Discord, with an admin role that overrides permissions and allows full control of the server with only the server owner able to assign and remove that role from other admins. As well, it should still be possible to remove roles from other users which are lower in the hierarchy than your current highest role, even if they have the same permissions as you. It is expected that your current highest role permission, as well as any roles above it, would be grayed out in the role assign menu for any user, indicating that you can't edit those roles, but you should still be able to assign and remove roles beneath your current one for everyone, including yourself.

The support server seems to believe this is a bug. I have tested this on the webapp in Firefox and Chrome, as well as the desktop app on Windows.

ItsSom1k:
Same thing happens as the owner of the server. Accidentally removed 'Admin' role from myself, and, suddenly, can no longer edit roles on anybody, myself included. Though I can change permissions on them just fine. Interestingly, I can edit the purely-visual roles, that do not grant any permissions, where as those that do are not available to me.

Experimented by creating a new role with complete permissions. This one I could add to myself just fine, but not the admin one.

mihaicm93:
https://github.com/stoatchat/for-web/pull/861 should fix a part of this
https://github.com/stoatchat/for-web/pull/865 should fix the other side

MidoriMushrooms:
> [#861](https://github.com/stoatchat/for-web/pull/861) should fix a part of this [#865](https://github.com/stoatchat/for-web/pull/865) should fix the other side

The issue is still persistent for me on both desktop and web. (Windows, Firefox.)

mihaicm93:
They are not yet into the public release, if you want them in now you will have to pull them in your own build

PR #865, opened 2026-02-21
fix: self role assignment

mihaicm93:
Users with permission to AssignRoles can now change / add / remove their own roles that are lower ranked then the current one

electron_mSnlluGMWR

Closes: https://github.com/stoatchat/for-web/issues/774

### DIFF
--- a/packages/client/components/app/menus/UserContextMenu.tsx
+++ b/packages/client/components/app/menus/UserContextMenu.tsx
@@ -288,9 +288,8 @@ export function UserContextMenu(props: {
           when={
             props.member?.server?.owner?.self ||
             (props.member?.server?.havePermission("AssignRoles") &&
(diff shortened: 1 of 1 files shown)


packages/client/components/app/menus/UserContextMenu.tsx
```diff
@@ -288,7 +288,9 @@ export function UserContextMenu(props: {
           when={
             props.member?.server?.owner?.self ||
             (props.member?.server?.havePermission("AssignRoles") &&
-              props.member.inferiorTo(props.member.server.member!))
+              props.member.inferiorTo(props.member.server.member!)) ||
+            (props.member?.server?.havePermission("AssignRoles") &&
+              props.user.self)
```
insertish: couldn't this just be a nested `inferior or self`?

PR #868, opened 2026-02-21
fix(file-input): reject unsupported image formats on upload

### LINKED ISSUE #859: Allows upload of invalid file format
### What happened?

When setting the server banner, it allows me to select and upload an `.avif` file despite when clicking save and exiting it does not allow it to be used. I feel there should either be a warning for invalid file format or not allow `.avif` from being uploaded.

Video of upload:

https://github.com/user-attachments/assets/e3444442-fd7c-4316-948a-2e4041e9d4be

DanMyers300:
# Summary
## [Fixes #859 ](https://github.com/stoatchat/for-web/issues/859)
The server banner (and other image file inputs) allowed .avif and other unsupported formats to be selected and submitted, despite the backend rejecting them.

# What changed
Form2.FileInput now validates the selected file's MIME type against the list of supported image formats (image/jpeg, image/png, image/gif, image/webp) before accepting it. If an unsupported format is chosen, an error modal is shown and the file is not set on the form control.
This is enforced in the shared Form2.FileInput wrapper, so it covers all image upload inputs across the app (server banners, user profile banners, server icons, channel icons, avatars, and webhook avatars).

# Testing
Try uploading an .avif file to the server banner, you should now see an error modal with the message "The file '...' has an unsupported format." and the banner should remain unchanged.

### DIFF
--- a/packages/client/components/ui/components/utils/Form2.tsx
+++ b/packages/client/components/ui/components/utils/Form2.tsx
@@ -160,7 +160,7 @@ const FormFileInput = (
             const file = files[0];
             if (!ALLOWED_IMAGE_TYPES.includes(file.type)) {
               const error = new Error(
-                t`The file "${file.name}" has an unsupported format.`,
+                t`The file "${file.name}" has an unsupported format. Supported formats are: JPEG, PNG, GIF, and WebP.`,
               );
               error.name = t`Unsupported file format`;
               openModal({ type: "error2", error });
--- a/packages/client/components/ui/components/utils/Form2.tsx
+++ b/packages/client/components/ui/components/utils/Form2.tsx
@@ -5,8 +5,8 @@ import {
   For,
   Match,
   Show,
-  Switch,
   splitProps,
+  Switch,
 } from "solid-js";
 
 import { Trans, useLingui } from "@lingui-solid/solid/macro";
@@ -15,7 +15,10 @@ import { css } from "styled-system/css";
 import { styled } from "styled-system/jsx";
 
 import { useModals } from "../../../modal";
-import { ALLOWED_IMAGE_TYPES } from "../../../state/stores/Draft";
+import {
+  ALLOWED_IMAGE_TYPE_LABELS,
+  ALLOWED_IMAGE_TYPES,
+} from "../../../state/stores/Draft";
 
 import { Button, Checkbox, Radio2, Text, TextField } from "../design";
 import { TextEditor2 } from "../features/texteditor/TextEditor2";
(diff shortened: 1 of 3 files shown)


DanMyers300:
https://github.com/user-attachments/assets/595af8d9-d169-4b72-ba78-36650c3b7433

gabrielfordevelopment:
Would be nice if you could give more information to the user, such as a list of the supported formats.
"File format is not supported. Supported formats: JPG, PNG....."
Information is really appreciated by users.

DanMyers300:
> Would be nice if you could give more information to the user, such as a list of the supported formats. "File format is not supported. Supported formats: JPG, PNG....." Information is really appreciated by users.

Thanks for the suggestion. I've included a message showing what file types are permitted.

image

gabrielfordevelopment:
> > Would be nice if you could give more information to the user, such as a list of the supported formats. "File format is not supported. Supported formats: JPG, PNG....." Information is really appreciated by users.
> 
> Thanks for the suggestion. I've included a message showing what file types are permitted.
> 
> image

In the long term, it might be better to **not** hardcode any of these.
Put them in a List instead and print out the list.
This way if Stoat will support more files later, it will dynamically extend the supported files List, so you we don't have to manually change the text every time

DanMyers300:
> In the long term, it might be better to **not** hardcode any of these.
> Put them in a List instead and print out the list.
> This way if Stoat will support more files later, it will dynamically extend the supported files List, so you we don't have to manually change the text every time

Good call! I've gone ahead and done exactly that. I added an ALLOWED_IMAGE_TYPE_LABELS array in Draft.ts right next to ALLOWED_IMAGE_TYPES, and the error message now builds the format list dynamically from it. I could programmatically derive the image labels from the image types list but I thought it would be nice to have an image label list as well in the future.

packages/client/components/ui/components/utils/Form2.tsx
```diff
@@ -148,8 +159,23 @@ const FormFileInput = (
...
+            const file = files[0];
+            if (!ALLOWED_IMAGE_TYPES.includes(file.type)) {
+              const formatList = ALLOWED_IMAGE_TYPE_LABELS.join(", ");
+              const error = new Error(
+                t`The file "${file.name}" has an unsupported format. Supported formats are: ${formatList}.`,
+              );
+              error.name = t`Unsupported file format`;
+              openModal({ type: "error2", error });
+              // FileInput resets to null before passing the new file for
+              // reactivity reasons, so restore the last accepted value
```
insertish, line 172: Maybe we can fix the root cause here instead of doing a workaround?

chrishultin: @DanMyers300 Any luck on this?

Dadadah:
This was fixed by #1422

PR #882, opened 2026-02-22
feat: Add push notifications to the service worker of the PWA

Dadadah:
Add push notifications to the service worker.

This PR adds the "Notifications" tab to the settings, with the sound options defaultly not shown.

You can click/tap on notifications to go to the message that notified.

An error modal will be shown if the user tries to enable notifications when denying notifications in their browser.

I could really use someone with an iPhone using Safari to help me test if this works. I have tested Android and desktop.

I've been receiving push notifications in my self-hosted environment for over 3 months now. 

This PR includes feature detection for desktop notifications. In practice, this doesn't change the new user experience. However, if a user wipes their cache but doesn't reset notification permissions, they may see a notification pop up for 100ms. This is required to disable the "Desktop Notifications" checkbox on mobile.

## Screenshots

<details>

New notification settings on web:
image
image

On for-web on mobile:
Screenshot_20260521-232650_Vivaldi

On desktop:
image

A notification on desktop (Gnome on Linux):
image

Note that the notifications will look different on Windows and Mac computers.

A notification on mobile (Android):
IMG_20260405_230251

Trying to enable notifications when you blocked them on the browser will create a modal:
image

Seldomly, enabling push notifications will fail. This will usually be due to too many requests or a configuration error. If this happens, a snackbar will be displayed to the user.
image

</details>

Gen AI was not used in the writing of this PR.


chrishultin (approved):

Pecacheu:
How's this one coming along now that all the upstream dependencies are taken care of? I'd love to see working push notifications, myself. I'm always missing messages on Stoat unless I check manually lol

Dadadah:
All the upstream dependencies have been taken care of, and this is working great in my self-hosted environment. However, there were some concerns brought up about scalability so I need to have a convo with the backend team to find out what to do about it.

Pecacheu:
Gotcha. Hope it gets figured out! Tbh my understanding was that push notifications were already _mostly_ implemented on the backend and it was mainly stoat.js and frontend work needed to make it happen, but idk that much about that to be fair lol

Dadadah:
This PR is now waiting on stoatchat/stoatchat#693. Once that is merged, and a new version is created on the backend, I can update this PR to use the new push notification structure and it should make it much simpler.

Misofist:
Sorry to nag you, but since stoatchat/stoatchat#693 has now landed, do you have an idea of when you might be able to work on this? I'm really looking forward to this feature.

Arsabutispik (approved):
LGTM

Pecacheu:
I can review & test this tomorrow

Pecacheu (changes requested):
We discussed in more detail in DMs but here's a summary for my suggestions:
1. Initial checkbox state should be off if permission is not accepted yet
2. If permission was accepted (which currently gets prompted to the user upon first-time login), push notifications should then default to on like desktop notifications do
3. Only show desktop notif checkbox if desktop notifs are supported *(Currently only desktop)*
4. Only show push notif checkbox is push notifs are supported *(Currently only mobile, for reasons beyond my mortal comprehension)*

packages/client/components/client/NotificationsWorker.tsx
```diff
@@ -192,18 +228,23 @@ export function NotificationsWorker() {
...
+    if (state.settings.getValue("notifications:desktop") === "default") {
+      Notification.requestPermission().then((permission) => {
+        if (permission === "denied") {
+          batch(() => {
+            state.settings.setValue("notifications:desktop", "denied");
+            state.settings.setValue("notifications:push", "denied");
+          });
+        } else {
+          state.settings.setValue("notifications:desktop", "allowed");
+        }
```
Pecacheu:
```suggestion
        const notifs = permission === "granted" ? "allowed" : "denied";
        batch(() => {
          state.settings.setValue("notifications:desktop", notifs);
          state.settings.setValue("notifications:push", notifs);
        });
```

This should address suggestions 1 and 2 above, and simplifies the code a bit. Here also I am treating "default" as a denial so that the app doesn't ask again until you manually try to enable notifications, even though clicking away from the prompt (eg. clicking the "X" instead of "Deny") isn't explicitly a no.

Pecacheu (changes requested):

packages/client/components/client/NotificationsWorker.tsx
Dadadah: Yes I agree, I will implement this, but your suggestion is a bit off so I'll do it on my branch.

Dadadah:
> We discussed in more detail in DMs but here's a summary for my suggestions:
> 
> 1. Initial checkbox state should be off if permission is not accepted yet
> 2. If permission was accepted (which currently gets prompted to the user upon first-time login), push notifications should then default to on like desktop notifications do
> 3. Only show desktop notif checkbox if desktop notifs are supported _(Currently only desktop)_
> 4. Only show push notif checkbox is push notifs are supported _(Currently only mobile, for reasons beyond my mortal comprehension)_

1 solved by your suggestion.

2 is not the desired behavior. Push notifications must be separately enabled.

I've added feature detection for 3. Please review my feature detection.

As for 4, most desktop browsers support push notifications. I do not know a way to detect push notification support.

edit: We spoke with @infi about this and determined that 2 was in fact the desired behavior. I've submitted a refactor that makes that the default behavior.

infi (approved):

Pecacheu (approved):

PR #884, opened 2026-02-23
feat: show file size limits where applicable

amycatgirl:
Should clear up some confusion about what the file size limits are across the app. Obviously, the app needs a way to get this from the backend (ideally via the root endpoint), but for now this shall suffice.

# Screenshots
image
image
image

### DIFF
--- a/packages/client/components/ui/components/utils/files/FileInput.tsx
+++ b/packages/client/components/ui/components/utils/files/FileInput.tsx
@@ -4,7 +4,6 @@ import { css } from "styled-system/css";
 import { styled } from "styled-system/jsx";
 
 import { ALLOWED_IMAGE_TYPES } from "@revolt/state";
-
 import { Button, Ripple } from "../../design";
 import { Row } from "../../layout";
 import { Symbol } from "../Symbol";
@@ -145,6 +144,7 @@ export function FileInput(props: Props) {
               <img src={imageSrc()} />
             </Show>
           </ImagePreview>
+
           <Show when={props.allowRemoval !== false}>
             <Button
               size="icon"
--- a/packages/client/components/app/interface/settings/channel/Overview.tsx
+++ b/packages/client/components/app/interface/settings/channel/Overview.tsx
@@ -106,7 +106,14 @@ export default function ChannelOverview(props: ChannelSettingsProps) {
           <Text class="label">
             <Trans>Channel Info</Trans>
           </Text>
-          <Form2.FileInput control={editGroup.controls.icon} accept="image/*" />
+          <Form2.FileInput
+            control={editGroup.controls.icon}
+            accept="image/*"
+            maxSize={
+              client().configuration?.features.limits.default
+                .file_upload_size_limits["icons"] ?? 2.5e6
+            }
+          />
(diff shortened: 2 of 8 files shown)


chrishultin:
Can we update this to pull the limits from the configuration?  https://github.com/stoatchat/stoatchat/pull/644 will expose that information, and https://github.com/stoatchat/javascript-client-sdk/pull/127 loads the configuration.

amycatgirl:
Sure! I would be glad to do so once both PRs are merged.

amycatgirl:
Seems like both PRs were merged, so I'll work on integrating that work into my branch

hopefully i don't get caught in conflict resolution hell again

amycatgirl:
hmm, i see that there isn't a type for the newly added `limits` field in the instance's configuration.

which means that the typescript linter/compiler will absolutely yell at me for trying to use it

should I make a PR to add said types?

amycatgirl:
also the ci bot **ruined** my branch again!!
how shameless...

chrishultin:
I'd love to see that!

amycatgirl:
ah, it seems that the API types are autogenerated, and according to the latest commit, it does include the `limits` field inside of `features`

im not sure if i have to point the client sdk to the latest commit though

chrishultin (dismissed):
This appears to still have hardcoded size limits - I thought the intent was to pull those dynamically?

amycatgirl:
yes, i was waiting on the js sdk submodule getting bumped because otherwise typescript complains about the `features.limits` object inside of `client.configuration` not existing

all i did was pull in changes from main

amycatgirl:
> "js sdk submodule getting bumped"

more specifically, the stoat-api package inside of the sdk, as that contains the types generated from the OpenAPI schema

amycatgirl:
i forgor :skull:

amycatgirl:
Rebased onto latest changes, and this PR is now using the instance configuration from the client.

packages/client/components/ui/components/utils/files/FileInput.tsx
```diff
@@ -36,6 +35,10 @@ interface Props {
    */
   accept?: "image/*";
 
+  /**
+   * How big the maximum size should be
+   */
+  max?: number;
```
Dadadah: Is this left over from a previous iteration? I don't think it's used.

amycatgirl: Yes sorry

amycatgirl: should be gone now

Dadadah (approved):

infi (approved):

Pecacheu:
Just a heads-up, this PR got merged but it is using the old way of getting config.

Instead of:
```
const client = useClient();

maxSize={
  client().configuration?.features.limits.default
    .file_upload_size_limits["avatars"] ?? 4e6
}
```

Should use:
```
const { limits } = useInstance();

maxSize={
  limits().file_upload_size_limits["avatars"]
  //No need for default value, configuration is always loaded within instance context
}
```

I believe we updated the docs to reflect this.

Issue #877, opened 2026-02-22
Zoomed-in browser view may hide thumbnails for attached images


SnoopJ:
### What happened?

Browser: Firefox 147.0.2
Stoat version: 0.2.1 (2025-10-10) running on `stoat.chat`

## Description

I have been using Stoat at a zoom level of 133% because of concerns similar to those identified in #64, and when I shared a message that had 4 image attachments (each image is 2992x2992px, in JPEG format), I noticed that the fourth image was cut off and not visible until I lowered my zoom back to 100%. See attached screenshot, note the rounded edges indicating the image that is almost entirely hidden by the message composition UI element.

<img width="1110" height="882" alt="Image" src="https://github.com/user-attachments/assets/07af2ad2-ec48-4dde-aa6f-b1dcf2e1e002" />

## Expected behavior

My main expectation is that Stoat would produce a grid of smaller image previews, taking up less vertical space in the chat history, but I do expect that messages will not be _hidden_ by a marginally-higher zoom level than 100%.

gaberashko:
I looked into this bug, and it seems that it's caused by a max-height: 200vh set on the message container element. Removing this rule fixes it entirely, prevents cutoff, and doesn't create any other undesirable overflow/overlap behavior.

Given that this seems like an arbitrary number, and doesn't account for several images existing in one container, I'm going to create a pull request with this rule removed. I don't see when it would be useful to enforce this height. In the event of multiple images/media, I feel as though a restriction would/should naturally exist on the number of images one can send in one message, or the total size being sent in one message.

SnoopJ:
> In the event of multiple images/media, I feel as though a restriction would/should naturally exist on the number of images one can send in one message, or the total size being sent in one message.

Agreed, this would make for a good issue.

> Given that this seems like an arbitrary number, and doesn't account for several images existing in one container, I'm going to create a pull request with this rule removed. I don't see when it would be useful to enforce this height. 

Cheers!

PR #885, opened 2026-02-23
fix: zoomed-in browser view hiding thumbnails

gaberashko:
removed max-height: 200vh causing thumbnails to overflow and be hidden in message container. preventing too many images from being sent should not be handled by arbitrary height enforcement.

fixes issue #877


insertish:
-snip-, I guess this is fine, I don't remember the reasoning for the max height but it doesn't make sense to keep it if we don't have a reason documented

PR #886, opened 2026-02-23
feat: Added toolbar for drafts and unsent messages

realRicFlair:
When replying to a recently unsent message, the draft message would not show a toolbar for unsending the message.
Only way to cancel a send was by right clicking, which was confusing.
I included the toolbar for draft messages with an unsend and retry button.

Also for messages that are in progress of being sent, I included a cancel button on the toolbar.

I also turned the delete button red. Not sure if I shouldv even done that tbh, but its super convenient and its like that on discord too, which people might expect. Although the color is off and idk if I shouldv set it to some variable instead.
image

### DIFF
--- a/packages/client/components/ui/components/features/messaging/elements/MessageToolbar.tsx
+++ b/packages/client/components/ui/components/features/messaging/elements/MessageToolbar.tsx
@@ -136,7 +136,7 @@ export function MessageToolbar(props: { message?: Message, draft?: UnsentMessage
       >
         <div class={tool()} onClick={deleteMessage}>
           <Ripple /> {/* I just set it to red cus it looks cool. Idk if i should. also maybe set it to some theme variable thing*/}
-          <MdDelete {...iconSize(20)} fill="red" />
+          <MdDelete {...iconSize(20)} fill="var(--md-sys-color-error)" />
         </div>
       </Show>
 


realRicFlair:
oops, meant to publicsh it as draft

realRicFlair:
alr i fixed the DCO check and fixed the pr title

realRicFlair (commented):
I checked through the code and it seems to work. No issues from what I see. Would be nice if someone reviewed the color change on the button though. I can remove that change if need be

packages/client/components/ui/components/features/messaging/elements/MessageToolbar.tsx
```diff
@@ -91,17 +110,38 @@ export function MessageToolbar(props: { message?: Message }) {
...
-          props.message?.channel?.havePermission("ManageMessages")
+          props.message?.channel?.havePermission("ManageMessages") ||
+          props.draft
         }
       >
         <div class={tool()} onClick={deleteMessage}>
-          <Ripple />
-          <MdDelete {...iconSize(20)} />
+          <Ripple /> {/* I just set it to red cus it looks cool. Idk if i should. also maybe set it to some theme variable thing*/}
+          <MdDelete {...iconSize(20)} fill="red" />
```
mihaicm93: Would be more fitting if you want it colored to have it using this
`          <MdDelete {...iconSize(20)} fill="var(--md-sys-color-error)" />`

insertish:
Can we have the screenshot updated? thanks!

PR #888, opened 2026-02-23
fix: Disabled submit button until input value contains discriminator

mihaicm93:
Disable `Add Friend` button if discriminator is not present in the expected form as `username#1234`

Should fix issues like -> Fixes: https://github.com/stoatchat/for-web/issues/1181

Before: 
firefox_2C6pIYj4Va

After: 
firefox_BkYeTLKyXK


insertish (changes requested):
Please drop `i18n/catalogs` from the PR, workflow has been changed, sorry for the inconvenience :slightly_smiling_face:

packages/client/components/modal/modals/AddFriend.tsx
```diff
@@ -47,18 +53,47 @@ export function AddFriendModal(
...
       isDisabled={group.isPending}
     >
       <form onSubmit={submit}>
-        <Form2.TextField
-          name="username"
-          control={group.controls.username}
-          label={t`Username`}
-          placeholder={t`username#1234`}
-        />
+        <div style={{ position: "relative" }}>
```
Dadadah, line 62: Form2 has built in error handling. If you add an error to `group.controls.username` on submit/validation, it should have an error string displayed, omitting the need for adding the error div below the form field.

Dadadah (changes requested):

PR #911, opened 2026-02-24
fix: Increase autocomplete box size

mihaicm93:
Set a maximum height for autocomplete box so emoji autocomplete box has more then 4 items visible

firefox_PxNfFV6emp


packages/client/components/ui/components/features/texteditor/codeMirrorAutoComplete.ts
```diff
@@ -27,6 +27,7 @@ const completionTheme = EditorView.theme({
   ".cm-tooltip.cm-tooltip-autocomplete > ul": {
     "min-width": "10em",
     "max-width": "50vw",
+    "max-height": "50vh",
```
infi: I am unsure about using 50% of the viewport height here as it means on very tall screens the autocomplete suggestion list would be extremely tall. Would it be possible to set a second upper bound at a different value (around 700px), whichever is smaller (`min()`)?

infi (changes requested):

packages/client/components/ui/components/features/texteditor/codeMirrorAutoComplete.ts
mihaicm93: I have adjusted it to 490px as 700px looked way to big

image

Dadadah (approved):

infi (approved):

PR #921, opened 2026-02-25
Update messages.po

Statium:
Translations into Russian


mihaicm93:
I might be wrong here but translations should be done on https://translate.stoat.chat/ , don't close this PR and wait for someone from team to confirm or deny my comment

insertish:
^ please use Weblate (also helps avoid things like the wrong field being changed -- msgid is changed here at least 3 times)

PR #923, opened 2026-02-25
fix: resolve voice chat ghost members and unknown user issues

SimonSeider:
## Summary

- Implement `VoiceChannelMove` event handler to properly remove users from source channel and add to destination channel when moving voice chats (fixes ghost members)
- Modify `useUsers` hook to fetch user profiles from API when not in cache, fixing "Unknown User" display issue when users join voice channels

### DIFF
--- a/packages/client/components/markdown/users.ts
+++ b/packages/client/components/markdown/users.ts
@@ -71,10 +71,19 @@ export function useUsers(
   // eslint-disable-next-line solid/reactivity
   return createMemo(() => {
     const client = clientAccessor()!;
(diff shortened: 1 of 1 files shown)


packages/stoat.js
amycatgirl, line 1: It looks like there was a change committed in the stoat.js submodule. The commit SHAs seem invalid, according to GitHub.

You should probably open a pull request on the [javascript-client-sdk](https://github.com/stoatchat/javascript-client-sdk) repo instead of directly committing to the submodule.

SimonSeider: Okay thanks, I will do that

PR #931, opened 2026-02-25
feat(client): add emoji picker to message edit box

youtsuhodev:
This PR adds an **emoji picker button** inside the inline message editing field (Discord-like), allowing users to insert emojis without typing `:` shortcuts. It also **hides the message hover toolbar** while editing to prevent UI overlap.

### Message Editing UI
- Added an **emoji button** inside the edit input container.
- Reused the existing `CompositionMediaPicker` / `EmojiPicker` from the message composer.
- Emoji insertion is handled via `TextEditor2` `nodeReplacement`, so it inserts at the current cursor selection.

### Hover Toolbar Behavior
- Disabled the message hover toolbar (`MessageToolbar`) while a message is being edited to avoid layering/overlap issues.

## Files Touched
- `packages/client/components/app/interface/channels/text/EditMessage.tsx`
- `packages/client/components/ui/components/features/messaging/elements/Container.tsx`

## Testing
1. Start editing a message.
2. Click the **emoji icon** inside the edit field and insert an emoji.
3. Type `:` in the edit field — emoji autocomplete should appear above the UI.
4. Hover the edited message — the hover toolbar should **not appear during editing**.

Signed-off-by: Lilian Vetlé <lilian.vetle@gmail.com>


Pecacheu:
Since there are no screenshots here to show what this looks like I took the liberty of taking one.


Personally I'm not a fan of adding a button there, but that's a subjective style question for the UI team. Additionally, this would have to be disabled on mobile because of the limited horizontal screen real-estate.

Issue #798, opened 2026-02-17
In Chrome, touch events on some buttons propagate through after dialogs close


Pecacheu:
### What happened?

This only occurs when using touch, not mouse. Tested in latest Brave Browser (Chromium fork) on my dev PC as well as latest stock Chrome for Android, but the issue does not seem to occur in Firefox for Android.

Here's a demonstration:
![Image](https://github.com/user-attachments/assets/a754e9a7-741d-424c-ae83-dc5de9ed5944)

I attempted to fix bug this myself, but I had to rollback the change because fixing the problem with `event.preventDefault()` caused issues on Firefox, as per [this discussion](https://github.com/stoatchat/for-web/pull/764#discussion_r2814776232) on my PR. `event.stopPropagation()` had no effect in my testing.

PR #666, opened 2026-02-12
fix: checkboxes doubble toggle, autostart and made toggle logic reusable

mihaicm93:
- Fix autostart by reading it asynchronously on mount and updating the stored value before updating UI state.
- Prevent checkbox/toggle double-click issues by stopping event propagation and reusing a shared toggle handler.

## Related Issues
- https://github.com/stoatchat/for-web/issues/643
- https://github.com/stoatchat/for-web/issues/581
- https://github.com/stoatchat/for-web/issues/667
- https://github.com/stoatchat/for-web/issues/679
- https://github.com/stoatchat/for-desktop/issues/10
- https://github.com/stoatchat/for-desktop/issues/14
- https://github.com/stoatchat/for-desktop/issues/44

## Notes
This PR only changes the web UI (Native.tsx). The desktop IPC change is already in a separate PR and is linked [here](https://github.com/stoatchat/for-desktop/pull/68) for context.
##
YWxit6yIEs

PR #937, opened 2026-02-26
fix: Checkboxs firing multiple onClick events when nested inside a CategoryButton

### LINKED ISSUE #647: Voice Processing checkbox not working
### What happened?

When trying to click the voice processing checkboxes they do not do anything, the element reacts but the checkbox does not change state, it only changes if clicked outside of the checkbox on the background element instead.

https://github.com/user-attachments/assets/38764ae6-8d57-4936-8439-f6bac5fb3412

silentsonata:
### Summary
This is my first PR so please let me know if something is incorrect and I will be happy to change it. I love the project and would love to help in any way I can.

### Bug
This PR fixes a bug when trying to select a checkbox when its nested inside of a CategoryButton from triggering the onClick function multiple times. An example of this is in the Voice settings. When pressing the check box on either of the processing options it appears to the user that the checkbox is not doing anything. What is happening is the function is called twice which means the option does not change. From the research that I could find on this behavior, it appears to be related to how the mdui-checkbox works and that it will trigger multiple onclicks under some situations. From further testing I found that this double trigger happens when the mouse presses on any element of the checkbox besides the <input> element.

### PR Explanation
To resolve the multiple trigger problem, I expanded the functionality of the onClick prop defined in the CategoryButton and use an Event to detect what the user presses on. The same logic is replicated so when the button is disabled the function does not run. Then I added logic to only allow the onClick function to execute if the user either pressed the CategoryButton (the <div> or <a>) or the <input> element on the checkbox.

### Test
I have confirmed that this fix does resolve the multiple triggers across all instances that there is a checkbox inside of the CategoryButton. The result is a single execution of the onClick function and the state of the checkbox properly updates no matter where the user clicks on the button.

Added by maintainers:
Fixes #647


mihaicm93:
I have pulled this changes as I already had some isolated fixes done to checkboxes ( https://github.com/stoatchat/for-web/pull/826 & https://github.com/stoatchat/for-web/pull/666 ) and this central fix looked promising at first but it seems like it breaks the click anywhere on the row now as seen below: 

electron_tL8mMQnC7i

Now clicking on chevrons ">" or icons gets ignored

silentsonata:
Interesting! Let me see if I can see what's causing that. My assumption is the row element has different events that wern't taken into consideration. Let me debug this and I'll get back to you with a solution.

silentsonata:
Just pushed an update that will only run the fix if a mdui-checkbox is present in the element. After testing all the buttons now work as expected with or without a checkbox.

mihaicm93:
I have pulled the latest changes to test, the issue from the Roles page now it's fixed but the checkboxes are still very inconsistent as seen below

electron_GzuyM0wnkE

packages/client/components/ui/components/design/CategoryButton.tsx
```diff
@@ -59,7 +59,38 @@ export function CategoryButton(props: Props) {
...
+        if (lastChild.tagName === "MDUI-CHECKBOX") {
+          if (
+            /**
+             * Only trigger onClick when clicking the button OR an input element such
+             * as a checkbox. This prevents multiple triggers of the onClick funciton
+             * due to mdui-checkbox calling two onClicks.
+             */
+            e.target instanceof HTMLDivElement ||
+            e.target instanceof HTMLInputElement ||
+            e.target instanceof HTMLAnchorElement
```
mihaicm93: I have looked into it and I see you don't target the SVG or the SPAN wich is causing the inconsistency, testing with addition of them fixes it
```
            e.target instanceof HTMLSpanElement ||
            e.target instanceof SVGElement
```

silentsonata: I ran your suggestion and that does appear to fix the problem for items with Icons but now it is broken again on the Voice page. I might need to rethink how I am preventing the double trigger. The original issue of double triggering appears again  when the HTMLSvgElement is found. Let me think and try some other ways out and I will post an update once I have something.

silentsonata:
Alright I completely rethought how to handle this and I think this is much cleaner and from my testing fixes all the issues that were mentioned. @mihaicm93 could you try this PR and see if it fixes the issues you reported?

mihaicm93 (approved):
Tested and all good

Fixes: https://github.com/stoatchat/for-web/issues/647

Superseds: https://github.com/stoatchat/for-web/pull/826

Pecacheu:
I haven't had a look at this up-close but do you think the same technique can be used to fix #798? I took a crack at it myself a while ago but my fix broke Firefox support while working with Chrome (on that note, make you y'all test this on **both** before merge cause it seems they handle click events differently O.o)

silentsonata:
@Pecacheu Just checked Firefox 149 and Chrome 147 and this fix does work on both platforms so we should be clear there.

Dadadah (dismissed):

Dadadah:
I think this can be done by doing just this:
```
onClick={(e: Event) => {
        // Prevent clicks when button is disabled
        if (props.disabled) return;

        e.preventDefault();
        props.onClick?.();
      }}
```

This seems to solve the issue in testing. Are you sure you need to actually check if the checkbox is being clicked?

silentsonata:
@Dadadah Thats a great question that I at one time had the answer to but can not remember off the top of my head. Let me update the branch and implement your solution and see if that resolves the issue as well.

silentsonata:
@Dadadah Thanks for the suggestion. This made me have a "duh" moment. From my testing this change works just fine and the Category Button now functions as expected.

Dadadah (approved):

Dadadah:
> @Dadadah Thats a great question that I at one time had the answer to but can not remember off the top of my head. Let me update the branch and implement your solution and see if that resolves the issue as well.

If you think of why feel free to message here, I will be testing this again in more depth this evening before merging it.

Dadadah:
OK works as expected on chromium based browser on android and desktop, and firefox on desktop.

Pecacheu (approved):

Pecacheu:
Tested and working on mobile/touch for me as well

PR #943, opened 2026-02-26
feat: add toggleable join and leave sounds for voice channels

Sadowski-Krystian:
Note: I temporarily placed the join/leave sound .ogg files in packages/client/src/audio/ so you can test this PR. I agree these should probably be global so the mobile apps can use them.


insertish (changes requested):
We are currently exploring sound branding & contacting people, so this PR will be blocked on that at least.

In regards to the functional changes here, this should instead introduce a sounds menu, much like in the old revite client, as we will be introducing other sounds in the app too. (i.e. being able to customise individual sounds being enabled/disabled)

Sadowski-Krystian:
I pushed a major update to address your feedback:

- Dedicated Sounds Menu: I've created a brand new "Sounds" tab in the client settings. It is built modularly, so when you are ready to introduce other app/notification sounds, they can easily be slotted in here.
- Individual Customization: Join and leave sounds have been split into individual, independent toggles (like in old Revolt).
- Custom Sound Uploads: Added a UI for users to upload their own custom sounds (saved locally as Base64 strings) to override the default ones.

I understand that this PR will remain blocked until the official sound branding is finalized. 

I have one question, right now, I left the state variables (`playJoinSound`, `customJoinSound`, etc.) inside the `Voice.ts` store. Since this new menu will eventually handle global app sounds, would you prefer I go ahead and extract this into dedicated `Sounds.ts` stores?

Sadowski-Krystian:
Sorry for this force push, forgot to run code formater. I'm new in open source contributing.

Dadadah:
Sounds were just merged in #1206

PR #946, opened 2026-02-26
fix: Display readable custom emoji names in notifications

### LINKED ISSUE #917: Use emoji name in notification instead of ID
### Describe your feature

When someone sends me an emoji in DM I only see the ID of the emoji in the notification, which doesn't tell me much.

Send the name of the emoji in notifications instead of the ID

Image

gabrielfordevelopment:
closes #917 
### Feature: Display readable custom emoji names in notifications

-   **Async Notification Processing:** Converted the onMessage handler in NotificationsWorker to an asynchronous function to support data fetching before displaying desktop notifications.
    
-   **Emoji ULID Extraction:** Added regex parsing to scan incoming message text (contentPlain) and extract all 26-character custom emoji IDs.
    
-   **On-Demand Emoji Fetching:** The client now checks the local cache for the extracted emoji IDs and automatically fetches any missing emojis from the API.
    
-   **Content Formatting:** Dynamically replaces raw emoji IDs with their human-readable names (e.g., `:cat:`) in the notification body. Includes a graceful fallback to a generic :emoji: string if an emoji cannot be resolved or fetched.

---

## Before:
image

---

## After:
image

### DIFF
--- a/packages/client/components/client/NotificationsWorker.tsx
+++ b/packages/client/components/client/NotificationsWorker.tsx
@@ -31,7 +31,7 @@ export function NotificationsWorker() {
    * Handle incoming messages
    * @param message Message
    */
-  function onMessage(message: Message) {
+  async function onMessage(message: Message) {
     const us = client().user!;
 
     // Ignore if we are currently looking at the channel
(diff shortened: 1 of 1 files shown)


Dadadah:
This should probably be done in stoat.js instead of in for-web. The notification calls stoat.js client.markdownToText which currently doesn't support emojis.

gabrielfordevelopment:
Looks like this got fixed in another PR so I'll close this one since it is not needed anymore

PR #947, opened 2026-02-26
fix: Adjust reaction badge padding and increase emoji size

gabrielfordevelopment:
-   **Reaction UI adjustments:** Decreased the vertical padding of the reaction badges (from var(--gap-md) to var(--gap-sm)) to make them look less bulky.
    
-   **Emoji scaling:** Increased the size of the emoji within the reaction badge from 1.2em to 1.4em for better visibility.
    
-   **Layout alignment:** Added alignItems: "center" to ReactionBase to ensure the enlarged emoji and the text count remain perfectly vertically centered.
    
-   **Add reaction button:** Tweaked the height of the "add reaction" button to closely match the new slimmed-down dimensions of the standard reaction badges.

---

## Before:
image

---

## After: 
image


infi:
Those are not changes that we want to explore at the moment.

gabrielfordevelopment:
Thanks for the feedback. I’d love for us to reconsider this, as the current implementation has some objective visual inconsistencies that I’m aiming to fix:

-   **Asymmetric Padding:** Currently, the top and bottom padding are uneven, which prevents the emoji from being centered. The combined vertical padding is actually equal to the height of the emoji itself, making the badge look unnecessarily bulky.
    
-   **Alignment Issues:** In the current version, the emoji sits slightly higher than the reaction count. My PR introduces `alignItems: "center"` to ensure they are perfectly aligned on the same baseline.
    
-   **Visibility:** I slightly increased the emoji size and slimmed the padding to give the UI a more balanced, professional feel.
    
## Currently we have this:
chrome_ndjCPx6DlV




## What I made have this:
chrome_frCPcuO59a


---

I’m more than happy to adjust the specific values (like the emoji scale or the exact padding size) if you have different preferences, but I believe the alignment and spacing issues definitely need addressing.

Could we take another look at this? I’m open to any suggestions on how to refine the approach.

---
*Edit: Just a quick note regarding the screenshots: the 'px' labels I used are for illustrative purposes to show the spacing ratio. Since I scaled the emoji, the actual values are relative, but I wanted to provide clear visual markers for the vertical padding inconsistency.*

gabrielfordevelopment:
Furthermore, right now the padding for the reaction badges are not consistent with other badge components, for example mentions:

image

Reaction badge currently has significantly larger padding on the bottom and top, compared to how the mention badge looks.

Which should be also fixed by this feature implementation on the PullRequest I made.

PR #948, opened 2026-02-26
fix: incorrect language format locking up app when blocked user types in chat

Jade3375:
fixes an issue causing the app to lockup when a blocked user has messages in currently active chat caused by the language formats being incorrect e.g. EN\_us -> EN-us

- `main` 
  - \#948 :point\_left:


chrishultin:
Generally looks OK.  Can you please squash your commits, and rebase this against `master` so we can make sure that everything still looks good?

Jade3375:
@chrishultin got that done now and it looks to still be working correctly

mihaicm93 (approved):

PR #953, opened 2026-02-28
feat: add automatic gain control to voice processing options

### LINKED ISSUE #563: Stoat messing with the Microphone Level
### What happened?

Hi, me and my brother noticed an Issue where everytime when Stoat Desktop is running, the microphone level in the windows settings is automatically lowered. This might be some auto-leveling that applies when we talk a bit louder, but often I dont notice the lowered level so my normal voice is too quiet and some parts even get cancelled out by the noise gate. I can even reset the level in the settings and watch it get lowered automatically when I talk a bit louder. This only happens when Stoat is running, and the issues also started appearing when we switched from Discord to Stoat. Unfortunaley, I couldnt find a setting in the app to prevent this behavior. 

Image
my normal microphone level

Image
the automatically adjusted level

Dadadah:
Closes #563 by allowing disabling of automatic gain control.

Many users (myself included) do not like automatic gain control. By default, livekit has it enabled. This PR adds a voice processing option to disable automatic gain control, with the default value being on.

### DIFF
--- a/packages/client/components/app/interface/settings/user/voice/VoiceProcessingOptions.tsx
+++ b/packages/client/components/app/interface/settings/user/voice/VoiceProcessingOptions.tsx
@@ -26,6 +26,15 @@ export function VoiceProcessingOptions() {
         >
           <Trans>Browser Echo Cancellation</Trans>
         </CategoryButton>
+        <CategoryButton
+          icon="blank"
+          action={<Checkbox checked={state.voice.autoGainControl} />}
+          onClick={() =>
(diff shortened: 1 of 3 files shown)


Wigldu:
Out of couriosity: What is missing here besides the reviews?
I'd really like this feature to be in Stoat.

The deploy error is a technical issue with token permissions?
And I dont get why the "display the branch..." fails. I can see no error.

Never participated in os projects before, but I'm interested in this.

mihaicm93 (approved):

insertish (approved):

PR #960, opened 2026-03-01
chore: add pnpm and nodejs to default.nix

Gtoasted:
When using nix, you typically do not install development packages like node system-wide, but instead install them in a devshell. I have added these two packages to make the devshell more complete.


insertish:
Tooling is provided through mise.

PR #985, opened 2026-03-03
fix: attachment width and colors

### LINKED ISSUE #984: Text embeds are unthemed and have incorrect width
### What happened?

Image

Relevant files in the codebase: https://github.com/stoatchat/for-web/blob/main/packages/client/components/ui/components/features/messaging/elements/TextEmbed.tsx

And potentially others.

matortheeternal:
Resolves #984.

### DIFF
--- a/packages/client/components/ui/components/features/messaging/elements/TextFile.tsx
+++ b/packages/client/components/ui/components/features/messaging/elements/TextFile.tsx
@@ -26,6 +26,10 @@ const Container = styled("pre", {
     flexDirection: "column",
     maxHeight: "120px",
     whiteSpace: "pre-wrap",
+
+    "& code": {
+      overflowWrap: "anywhere",
+    },
   },
 });
 


matortheeternal:
these screenshots show updated attachment style provided by this PR.

## before
image

image

IMG_1470

## after
image

image

Pecacheu:
For the music player one, that dark background color looks sort of awkward, is it possible to go from the left one to something like on the right? Alternatively if the dark background went edge-to-edge at the bottom with no border, it would look better.

matortheeternal:
> For the music player one, that dark background color looks sort of awkward, is it possible to go from the left one to something like on the right? Alternatively if the dark background went edge-to-edge at the bottom with no border, it would look better.
> 

that's the `<audio>` element, and is browser-controlled.  ::-webkit-media-controls-enclosure exists, but is not universal.  generally speaking I would recommend against attempting to style this element.  a custom audio player is possible but is out of scope for this PR.

Pecacheu:
Oof, that's unfortunate. I really wish the W3C spec included more requirements on being able to theme built in browser elements instead of having to build your own alternative from scratch if you want any amount of customization.

Pecacheu:
Okay I had some time to actually test this PR. Definitely looks better than them going all the way to the edge with wasted blank space, but now the widths are inconsistent depending on the file type and length of the name, which looks a bit wonky. Also I noticed that the download link has a weird underline on hover (because the button is wrapped in an `a` element.)



So I made a few adjustments to the CSS, prevented the filename from overflowing with `overflow: hidden` and `text-overflow: ellipsis`, and added `max-width: 100%` so that the attachments scale down for mobile. Here's the result:




_(Side note, not sure how I can send the commit to you, I just became a maintainer lol. I do see the 'Maintainers are allowed to edit this pull request.' but I'm not yet sure how that feature works, dunno if I can just push to your PR's branch or what.)_

matortheeternal:
@Pecacheu personally, I don't think variable widths is a "problem".  To me the second example shown here looks worse with the exception of the final example which overflowed the available width.  I do think not growing beyond the bounds of the parent container (flexbox `min-width: 0;` situation?) is important though.

generally speaking I don't think I'd want a direct commit on my branch without talking with me first when the commit is not just to fix a bug/problem.  if you want to take over the PR or fold its changes into a larger PR you can, but I think that would limit my capacity and interest in contribution.

regarding the <a> button - I did not change that, it was a pre-existing choice.  my goal in this PR is to maintain a limited scope.  continuously adding more things to the scope makes it harder for the PR to be merged and increases the maintenance required to keep the PR alive.  If there is a desire for a more complete overhaul of attachments and embeds, I would want to have a proper conversation with some maintainers to get a better sense of what that would look like and how it should happen.

Pecacheu:
> personally, I don't think variable widths is a "problem". To me the second example shown here looks worse with the exception of the final example which overflowed the available width.

I mean, that's an... opinion. I suppose style is subjective, but on the other hand, we also don't want to create theming incontinency with what's already there, and embeds already enforce a consistent width, eg:


You can even send a message with both embeds and attachments on it at once, so it'd be kinda strange for one to enforce a consistent width and the other to just kinda go as long or short as it feels like.

> generally speaking I don't think I'd want a direct commit on my branch without talking with me first when the commit is not just to fix a bug/problem.

I totally agree, force-pushing commits to someone else's PR would be weird, I'm just saying I don't know how else to do it atm. Ideally there should be some sort of "commit suggestion" feature. Not sure if there's a way to do that.

> regarding the button - I did not change that, it was a pre-existing choice

Not sure when that happened then, cause I'm not seeing the underline in the production app, but my test one has multiple PRs shoved into it atm, so it might be another one breaking it lol. I'll keep that change separate.

matortheeternal:
> I mean, that's an... opinion.

I think there's a pretty good argument for it though.  As it stands, attachments will fill all available width.  I'm pretty sure that is not desirable, as the content itself is pretty much always of a fixed, known width.

Consider ultrawide monitors.  Is this really a sensible style to use?

image

My personal gut sense of this is no: attachments should use only the width they need to display their content.

> You can even send a message with both embeds and attachments on it at once

And media embeds can have distinct widths, because media has different built-in aspect ratios.

Let's consider the options:

- Option 1: Force everything to 100% (current behavior)
  Key disadvantages: noisy in the UI, space doesn't serve any function, the download button is farther away from the content, looks bad on UW monitors, is unnecessary.
  Key advantages: trivial consistency with everything else that uses 100% width.
- Option 2: Fixed width
  Key disadvantages: doesn't use available space when dealing with larger blocks of text, is still inconsistent with arbitrary aspect ratio media embeds, will often be "too small", causing wrapping and reducing readability.
  Key advantages: Predictable tiling, predictable sizing, consistency.
-  Option 3: Dynamic width:
  Key disadvantages: Not predictable sizing.
  Key advantages: Dynamic tiling is still possible, sizing is always appropriate for content, UX for downloading is generally good (cursor doesn't have to move as far to reach the download button), plays nicely with UW monitor setups.  Chat messages already have dynamic width, with a mixtures of GIFs, text, and other media.
  
There is a degree of subjectivity to style, but there are measurable pros and cons to different approaches. Generally my thinking is to look at what other software does, evaluate its efficacy, assess real user experience, and utilize well-understood UX principles.  Based on these I came to the conclusion that dynamic-width attachment previews gives the best UX.  So while my conclusion is subjective, the things I considered to come to it are not.  I think "tiling" is a separate discussion and outside of the scope of this PR.  There are many possible approaches to tiling, and it's really its own thing.

If we can't agree however, then I think it could also be a possible case for a user-controlled Display option.

> Not sure when that happened then, cause I'm not seeing the underline in the production app, but my test one has multiple PRs shoved into it atm, so it might be another one breaking it lol. I'll keep that change separate.

It's the current behavior of Stoat v0.2.1/Stoat for Desktop 1.3.0.

Pecacheu:
It is definitely subjective, it's a style thing, I'm just saying that being consistent with other aspects of the app's already-chosen design philosophy is important. Eg. If we're going to make these dynamically sized we also have to make embeds dynamically sized or else it looks inconsistent. I'm not saying that that's impossible, we could do that too, tho personally I much prefer the consistent sizing on both fronts.

Either way we can at least agree that Option 1 (current attachment behavior) is the worst one lol.

Tho brining up what other apps are doing is also a good point, so I decided to see how Discord and Telegram display those files. It looks like Discord (left) displays everything the same size *except* for text files for some reason, and Telegram (right) displays everything the same size *except* for music.




I guess since I don't send music or text files often I just always figured they're always consistently-sized for all attachments, so interesting result, but I'd still argue that it's more consistently sized than completely dynamic, it's more just that they have singular exceptions to the rule.

matortheeternal:
> I'm just saying that being consistent with other aspects of the app's already-chosen design philosophy is important. Eg. If we're going to make these dynamically sized we also have to make embeds dynamically sized or else it looks inconsistent.

I don't think I agree.  Website embeds are a wholly different beast because websites come with metadata that can be used to render a description, thumbnail, and title for a given embed.  The way this metadata works is standardized, so we can make a fixed width "tile" for a website embed and have it make sense.

Attachments are different.  Attachments do not come with preview metadata.  They also are not generally viewable in a web interface without being downloaded, so having a well-functioning preview is important.

Discord uses variable width for previewing text attachments, up to a maximum width which is actually <100% of the available space.  Discord text attachments do not have a minimum width, so they can be considered dynamically sized.  Code blocks use the same logic. 

Attachments which do not have a preview and music embeds have a fixed width.  This seems sensible, as the only variable content is the title.

All website embeds have the same fixed width, with variable height.

Telegram doesn't display previews of text content, so it is not comparable.

Based on this, I think I'm on board with having a fixed width for non-text preview attachments which is not 100% of the available width.

matortheeternal:
Consistent widths implemented:

image

image

Pecacheu:
Okay this one seems a lot better. I still prefer everything being the same width, however the text containers least do have a use for that extra space, so it makes sense from a UI perspective why they'd be wider than the others and doesn't hurt design consistency as much compared to everything being dynamically sized.

Would appreciate comments from the other maintainers on this but for me this seems like an acceptable approach.

A few minor issues remained from my test tho:
- If line in the file is too wide, but doesn't contain any spaces to break on, it doesn't break and creates a vertical scrollbar. This can be fixed with overflowWrap, I'll see if I can submit a code suggestion, finally figured out how to do those lol.
- The dynamic width containers can overflow past the edge of the screen. This is because even though max width is 100%, min width is given in explicit pixels instead of percent, so it overrides the max when there's a conflict.
- For me the audio element doesn't stretch to the entire width of the attachment box. This might be browser specific since audio elements vary by browser, but adding a `width: 100%` to it fixed it for me.

In other news I found the code suggestion feature so I'll see if I can get it to work lol

packages/client/components/ui/components/features/messaging/elements/FileInfo.tsx
```diff
@@ -22,6 +22,14 @@ const Base = styled(Row, {
   base: {},
 });
```
Pecacheu, line 23:
```suggestion
```

Not needed cause this is just Row lol

packages/client/components/ui/components/features/messaging/elements/FileInfo.tsx
Pecacheu, line 94:
```suggestion
    </Row>
```

packages/client/components/ui/components/features/messaging/elements/Attachment.tsx
```diff
@@ -19,8 +19,23 @@ export const AttachmentContainer = styled(Column, {
...
+
+export const DynamicAttachmentContainer = styled(Column, {
+  base: {
+    padding: "var(--gap-md)",
+    borderRadius: "var(--borderRadius-md)",
+    color: "var(--md-sys-color-on-secondary-container)",
+    background: "var(--md-sys-color-secondary-container)",
+    alignSelf: "start",
+    minWidth: "420px",
+    maxWidth: "100%",
```
Pecacheu, line 38:
```suggestion
export const AttachmentContainer = styled(Column, {
  base: {
    padding: "var(--gap-md)",
    borderRadius: "var(--borderRadius-md)",
    color: "var(--md-sys-color-on-secondary-container)",
    background: "var(--md-sys-color-secondary-container)",
    alignSelf: "start",
    width: "420px",
    maxWidth: "100%",
  },
  variants: {
    dynamic: {
      true: {
        width: "auto",
        minWidth: "min(420px, 100%)",
      },
    },
  },
});
```

Using the CSS min operation seems to be the best way to fix the clipping from max/min width conflict. Also, using the Panda variants feature makes the code a bit cleaner.

packages/client/components/ui/components/features/messaging/elements/Attachment.tsx
```diff
@@ -103,12 +113,10 @@ export function Attachment(props: { file: File; message?: Message }) {
...
       <Match when={props.file.metadata.type === "Text"}>
-        <AttachmentContainer>
+        <DynamicAttachmentContainer>
           <FileInfo file={props.file} />
-          <SizedContent width={480} height={120}>
-            <TextFile file={props.file} />
-          </SizedContent>
-        </AttachmentContainer>
+          <TextFile file={props.file} />
+        </DynamicAttachmentContainer>
```
Pecacheu, line 119:
```suggestion
        <AttachmentContainer dynamic>
          <FileInfo file={props.file} />
          <TextFile file={props.file} />
        </AttachmentContainer>
```

packages/client/components/ui/components/features/messaging/elements/FileInfo.tsx
Pecacheu, line 50:
```suggestion
    <Row align>
```

packages/client/components/ui/components/features/messaging/elements/TextFile.tsx
```diff
@@ -24,6 +24,8 @@ const Container = styled("pre", {
     overflow: "auto",
     scrollbarWidth: "thin",
     flexDirection: "column",
+    maxHeight: "120px",
+    whiteSpace: "pre-wrap",
   },
 });
```
Pecacheu, line 34:
```suggestion

    "& code": {
      overflowWrap: "anywhere",
    },
  },
});
```

Overflow wrap for text container

Pecacheu:
Okay so in conclusion the suggested change feature is slightly broken and doesn't render changed lines correctly but you should get the jist at least lol

Pecacheu:
I can see not wanting to change the Base -> Row thing since it was preexisting (even though it's an easy fix in the same spot as your modified code), but the Panda variants and CSS min op are legitimate directly related fixes and you gave no reason for rejecting them.

The former brings consistency with how we are using Panda CSS elsewhere, and without the latter, it's still possible for dynamic width containers to overflow the screen width.

PR #989, opened 2026-03-03
feat: quick access user profile and DM in context menu

### LINKED ISSUE #286: feat: quick access user profile and DM from user context menu

### LINKED ISSUE #1073: UI/UX: Improve context menu consistency for destructive actions
### Describe your feature

The context menu currently highlights "Report User" in red, but "Remove Friend" and "Block User" use the default text color.

1. We should make all destructive actions red for UI consistency and to prevent accidental clicks.

Image

2. Menu Reorganization: Rearrange the context menu items to logically group related actions together. Incorporating the upcoming "Profile" option, the top section of the menu should be ordered as follows:
-   Profile (opens user profile - referencing the existing open issue #286 with the implementation at #989 )
-   Message
-   Close Chat
    
The mute/notification settings would follow, and all red destructive actions should be logically grouped at the very bottom.

gabrielfordevelopment:
Closes #286 I believe
Closes #1073 
*I had to close the other PR cause I accidentally pulled an earlier bugged version to my own*

---

- Added "Profile" button to user context menu, always visible. Opening is blocked at handler level when the profile modal is already open, instead of hiding the button.
- Reorganized user conctext menu items into clearly separated categories with dividers: Quick Actions, DM-specific, Server Identity/Roles, Social, Moderation, Safety, and Developer tools.
- Added `destructive` flag to Block user, Remove friend, Reject friend request, Cancel friend request, and Close chat actions.
- Extracted permission checks into helper functions (`canEditIdentity`, `canEditRoles`, `canKick`, `canBan`, `canBanNonMember`).
- Added "Copy text" option to the draft message context menu. (It helps **a lot** since you can't edit it)
- Added a divider in the draft message context menu to separate non-destructive actions from the destructive "Delete message" action.
- Import order changed because of `prettier` format config.

## Server (Current: Left side - - What I made: Right side)
ServerClickb

## Direct Message (Current: Left side - - What I made: Right side)
DirectMessageb

## Server Member Panel (Current: Left side - - What I made: Right side)
ServerMemberPanelb

## Draft Message (Current: Left side - - What I made: Right side)
DraftMessageb

## Voice Channel (Current: Left side - - What I made: Right side)
VoiceChannelb

### DIFF
--- a/packages/client/components/app/menus/UserContextMenu.tsx
+++ b/packages/client/components/app/menus/UserContextMenu.tsx
@@ -1,15 +1,5 @@
-import { JSX, Match, Show, Switch } from "solid-js";
-
 import { Trans } from "@lingui-solid/solid/macro";
-import { useNavigate } from "@solidjs/router";
-import { Channel, Message, ServerMember, User } from "stoat.js";
-
-import { useClient } from "@revolt/client";
-import { useModals } from "@revolt/modal";
-import { useSmartParams } from "@revolt/routing";
-import { useState } from "@revolt/state";
-import { Slider, Text } from "@revolt/ui";
-
+import MdAccountCircle from "@material-design-icons/svg/outlined/account_circle.svg?component-solid";
 import MdAddCircleOutline from "@material-design-icons/svg/outlined/add_circle_outline.svg?component-solid";
 import MdAdminPanelSettings from "@material-design-icons/svg/outlined/admin_panel_settings.svg?component-solid";
 import MdAlternateEmail from "@material-design-icons/svg/outlined/alternate_email.svg?component-solid";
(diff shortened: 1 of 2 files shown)


Pecacheu:
_(Just commented this on the old one you closed)_

Think we could get that "Profile" button in the DM list context menu as well? Otherwise there's currently _no way_ to open the profile of a friend who you just added but whom hasn't DMd you back yet.

gabrielfordevelopment:
*Deleted comment - outdated screenshot. Check first comment to see current state*

Pecacheu:
Ok so I was giving this a test and it mostly works, but the "Message" option in the menu doesn't function. I get this error in the console:


Seems the problem is that it's trying to navigate to the full URL instead of only the path, which will fail on anything but the main instance due to the domain not matching. So you need to change line 64 in **openDm()** like so:
```
props.user.openDM().then((channel) => navigate(channel.path));
```

EDIT: Wait nvm apparently you didn't even change that line, and there's also an identical line that needs to be changed in `ProfileActions.tsx`, which you didn't even touch... So uh, yeah I guess this has just been broken on self-hosted instances the whole time and nobody noticed. Wild. We can say that's the scope for another PR then.

packages/client/components/app/menus/UserContextMenu.tsx
```diff
@@ -73,6 +74,16 @@ export function UserContextMenu(props: {
...
 
+  /**
+   * Open user profile
+   */
+  function openProfile() {
+    openModal({
+      type: "user_profile",
+      user: props.user,
+    });
+  }
```
Pecacheu, line 99:
```suggestion
    dismissFloatingElements();
  }
```

If this is triggered from User Card preview, the preview should be dismissed when opening the full profile, as it normally does when clicking the header of the User Card preview to open full profile.

packages/client/components/app/menus/UserContextMenu.tsx
```diff
@@ -10,6 +10,7 @@ import { useSmartParams } from "@revolt/routing";
 import { useState } from "@revolt/state";
 import { Slider, Text } from "@revolt/ui";
```
Pecacheu:
```suggestion
import { dismissFloatingElements, Slider, Text } from "@revolt/ui";
```

Import for below change

Pecacheu (changes requested):

gabrielfordevelopment:
> Review + change request by @Pecacheu 


- Added dismissFloatingElements as suggested in comments
- 1. *(This is good because opening full profile from the context menu now matches the existing User Card behavior and reliably closes the preview floating state first)*
- Updated screenshots
- Fixed DCO check

Pecacheu (approved):

gabrielfordevelopment:
Put it back to `Draft`. Will rework it to be more consistent across all tooltips

gabrielfordevelopment:
Alright, it is done now

Dadadah (approved):

mihaicm93 (approved):

infi (approved):

Issue #675, opened 2026-02-12
Emote File Selector MIME types incorrect


mm12:
### What happened?

The MIME type that tells the file selector what extensions to allow is incorrect. For example, it allows `svg` files, even though they do not work with stoat. 

The MIME types should be changed, and/or the additional file types should have support added.

PR #990, opened 2026-03-03
feat: configurable Form validation + EmojiList Integration

### LINKED ISSUE #806: Error upon attempted creation of some emojis
### What happened?

Trying out Stoat as a Discord alternative for my server incase the new age verification makes moderation impossible, and it keeps giving me this error when I attempt to put some images in:

Image

I'll also attack a file incase the paste didn't work which I doubt it did

Image

### LINKED ISSUE #1080: can't add emoji with capital letters in name, confusing error message.
### What happened?

when trying to add an emoji i did not know that caps weren't allowed, and the error message did not help me understanding what was wrong at all. Only trial and error led me to the conclusion that caps aren't allowed.

If this limitation is intended, please make the error message clear or display the naming rules next to the text field. If the emojis are meant to support capital letters, then this is actually a bug.

How to reproduce : 
1. in a server where you have the perms to do so, add a new emoji
2. give it a name that contains at least one capital letter
3. a regex parsing error comes up

example : 

Image

mihaicm93:
- Addition of configurable validators in Form2 for TextField (pattern, patternError) and FileInput (types, typeError, maxSize and sizeError) so they can be reused in other form fields

- Updated EmojiList to use the new form validation with direct feedback as seen below:
firefox_ke5IayrTjC

---
Closes: https://github.com/stoatchat/for-web/issues/806 https://github.com/stoatchat/for-web/issues/1080 https://github.com/stoatchat/stoat.chat/issues/99 and (after latest commit) https://github.com/stoatchat/for-web/issues/675 from what I can see

---
Server banner/Server Icon/Profile and other places that I can't think about, will still keep the previous working state (where any image type can be selected but now after latest commits they will get filtered out ) till patching them at a later date


packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
```diff
@@ -28,6 +29,16 @@ export function EmojiList(props: { server: Server }) {
   const client = useClient();
   const { openModal } = useModals();
 
+  // Validation requirements
+  const namePattern = /^[a-z0-9_]{1,32}$/;
+  const imgTypes = ["image/png", "image/jpeg", "image/gif", "image/webp"];
+  const maxSize = 500 * 1024;
```
chrishultin, line 34: Can we pull this from the configuration now?

chrishultin (changes requested):

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
mihaicm93: I can not figure out how

mihaicm93: > I can not figure out how

no wonder, as it's not avalible yet, so I guess would need to be revisited at a later date

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
```diff
@@ -28,6 +29,16 @@ export function EmojiList(props: { server: Server }) {
   const client = useClient();
   const { openModal } = useModals();
 
+  // Validation requirements
+  const namePattern = /^[a-z0-9_]{1,32}$/;
+  const imgTypes = ["image/png", "image/jpeg", "image/gif", "image/webp"];
```
insertish: Pretty sure we define these as constants somewhere already.

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
```diff
@@ -28,6 +29,16 @@ export function EmojiList(props: { server: Server }) {
...
+  const maxSize = 500 * 1024;
+
+  // Validation errors
+  const namePatternError = t`Lowercase, numbers and underscores only.`;
+  const invalidTypeError = t`File type not supported, only ${imgTypes} are allowed.`;
+  const imageSizeError = t`Image size can be maximum ${maxSize / 1024}KB`;
```
insertish, line 38: This now also duplicates invalid type error from another PR, although not merged, we should probably consolidate.

insertish (changes requested):
I think this is a dupe of https://github.com/stoatchat/for-web/pull/868 which fixes validation more broadly but is only scoped to file types.

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
mihaicm93: replaced imgTypes it with the file types given in state/store/Draft, I did not find anywhere the regex for namePattern

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
mihaicm93: I have moved the invalid type error inside of the Form2.FileInput so it gets served golbally from there

Dadadah:
This was superseded by #1423

Issue #939, opened 2026-02-26
Cross-origin requests to Stoat upstream servers when self-hosting


sinus-x:
### What happened?

When running the `for-web:addb6b7` image, along with `:v0.11.1` versions of backend services, requests are being made to `geo.revolt.chat` every few seconds, and four asset ones to `static.stoat.chat` upon opening the emoji picker. All these requests are being denied through CORS.

Each emoji asset request also emits "Cookie “” has been rejected as third-party." info message due to the mismatched domains.

PR #993, opened 2026-03-04
fix: use server config for custom emoji URLs instead of hardcoded CDN

### LINKED ISSUE #561: Move emoji and geo URLs to config
Right now the URLs for emoji and geo/AgeGate verification are hardcoded, which prevents overriding them for dev or self-hosted.

https://github.com/stoatchat/for-web/blob/e40d30c944fc23ec9c162348af44178a4be2b6bb/packages/client/src/interface/channels/AgeGate.tsx#L37-L43

https://github.com/stoatchat/for-web/blob/e40d30c944fc23ec9c162348af44178a4be2b6bb/packages/client/components/markdown/emoji/UnicodeEmoji.tsx#L56-L61

sanasol:
Custom emoji URLs are hardcoded to `cdn.revoltusercontent.com/emojis/` in the ProseMirror model builder and TextEditor component. This breaks on any instance using a different Autumn CDN URL.

Uses `client.configuration.features.autumn.url` from server config instead.

Fixes #561 (emoji part), partially addresses #939 (custom emoji cross-origin)

Tested and live on https://chat.sanhost.net

### DIFF
--- a/packages/client/components/markdown/prosemirror/to-model.ts
+++ b/packages/client/components/markdown/prosemirror/to-model.ts
@@ -147,7 +147,7 @@ function map(
     case "customEmoji":
       return schema.nodes.rfm_custom_emoji.createAndFill({
         id: node.id,
-        src: `https://cdn.revoltusercontent.com/emojis/${node.id}`,
+        src: `${client.configuration?.features.autumn.url}/emojis/${node.id}`,
       })!;
     case "unicodeEmoji":
       return schema.nodes.rfm_unicode_emoji.createAndFill({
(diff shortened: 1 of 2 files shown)


Dadadah (commented):
The TextEditor component is not used anywhere, I believe it is deprecated code. The to-model.ts file is only imported in the TextEditor. Therefore this code doesn't change what URL's for-web uses. The issues you referenced were referencing the emoji packs, not custom emojis, which are already handled by autumn. The emoji packs are hosted by a cdn, not the official Autumn. If you replace the cdn with Autumn, your emojis would fail to load.

PR #996, opened 2026-03-04
feat: update GIF embed detection from Tenor to Giphy URLs

sanasol:
Updates GIF embed detection to match Giphy URLs instead of Tenor URLs, since the gifbox backend is migrating from Tenor to Giphy.

Changes `https://tenor.com` → `https://giphy.com` in:
- `Embed.tsx` (embed component GIF detection)
- `Message.tsx` (inline GIF message detection)

Related: stoatchat/stoatchat#661, stoatchat/for-android#94

Tested and live on https://chat.sanhost.net

### DIFF
--- a/packages/client/components/app/interface/channels/text/Message.tsx
+++ b/packages/client/components/app/interface/channels/text/Message.tsx
@@ -85,7 +85,7 @@ export function Message(props: Props) {
     props.message.embeds[0].type === "Website" &&
     ((props.message.embeds[0] as WebsiteEmbed).specialContent?.type === "GIF" ||
       (props.message.embeds[0] as WebsiteEmbed).originalUrl?.startsWith(
-        "https://tenor.com",
+        "https://giphy.com",
       )) &&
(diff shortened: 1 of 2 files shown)


Pecacheu:
Interestingly for me, Giphy URLs already work with embeds exactly as Tenor GIFs do. I think it might be because it's passing this first check `(props.embed as WebsiteEmbed).specialContent?.type === "GIF"`, thus bypassing the second check of the domain name.

IAmTomahawkx:
The gifbox backend is not moving to giphy, unsure where that notion came from.

sanasol:
@IAmTomahawkx sorry for misleading, it should come together with https://github.com/stoatchat/stoatchat/pull/661
But thats just proposal/for other users who searching alternative gif provider since google is not providing new api keys for tenor now.

Pecacheu:
I suppose for now we should close this then, until the alternative that Zomatree mentioned is ready to be switched to.

PR #997, opened 2026-03-05
refactor: place instance-specific config into dedicated context

Gtoasted:
In order to implement token federation (as per https://github.com/orgs/stoatchat/discussions/1073#discussioncomment-15814999), the client must be refactored to not use global configuration for values that are dependent on a specific instance.
This is fairly easy for most of them, since they are accessed in components and can thus use solid's context feature, but some values are also accessed in stores. I'm not sure how to best make those access the config, and I would appreciate suggestions, but I believe they will lead to a few client errors at worst, so it should be fine to deal with them later.


Gtoasted:
These changes are already merged in #999, so to reduce pr clutter I'll close this one.

PR #999, opened 2026-03-06
feat: Multi-account switching & self-hosted/on-premise instances

Pecacheu:
# Dependencies
- #1217
- #1315

# Related Discussions
- https://github.com/stoatchat/self-hosted/issues/141
- https://github.com/stoatchat/for-desktop/issues/43
- https://github.com/stoatchat/for-desktop/pull/55
- https://github.com/stoatchat/for-web/issues/851
- https://github.com/stoatchat/stoatchat/issues/471
- https://github.com/orgs/stoatchat/discussions/1073#discussioncomment-15783714
- https://github.com/orgs/stoatchat/discussions/1310

# Current UI concept

Test this live at https://forestfire.net:420

_Feel free to give it a test if you've got a self-hosted instance and let me know how it goes!_

PR #1002, opened 2026-03-07
feat: access other instances

Gtoasted:
Building on #997, I am currently implementing token federation. Currently, it is possible to prepend any path with /instance/hostname to instruct the client to access the instance at https://hostname, provided the api is located at https://hostname/api. My plan is to wrap all links in the instance.href function, so that they point to the path with the correct instance prefix. This should provide a basis to implement things like token federation or account switching across instances, as per #999.
As always, feedback would be greatly appreciated.

### DIFF
--- a/packages/client/components/app/interface/settings/channel/Overview.tsx
+++ b/packages/client/components/app/interface/settings/channel/Overview.tsx
@@ -5,10 +5,10 @@ import { Trans, useLingui } from "@lingui-solid/solid/macro";
 import type { API } from "stoat.js";
 
 import { useClient } from "@revolt/client";
-import { CONFIGURATION } from "@revolt/common";
 import { useModals } from "@revolt/modal";
 import { Button, CircularProgress, Column, Form2, Row, Text } from "@revolt/ui";
 
+import { useInstance } from "@revolt/instance";
 import { ChannelSettingsProps } from "../ChannelSettings";
 
 /**
@@ -18,6 +18,7 @@ export default function ChannelOverview(props: ChannelSettingsProps) {
   const { t } = useLingui();
   const client = useClient();
   const { openModal } = useModals();
+  const instance = useInstance();
 
   /* eslint-disable solid/reactivity */
   // we want to take the initial value only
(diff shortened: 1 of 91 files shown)


Pecacheu:
Tbh I think this could be part of #997 to reduce the PR clutter a little. Good idea for how to handle it tho, but I think `/instance/hostname` might be a bit long (since it'll be visible in the URL bar shorter is better), how about just `/i/hostname`. We'd also need to deal with alternative ports somehow, not sure if `:` is allowed in the URL (it might be, I'd have to double-check)

Also btw I made some changes to Instance.ts in my PR here.
https://github.com/stoatchat/for-web/pull/999/changes/dd7304841f95041b0a0c72b2e6cdfabf61f4a3f7#diff-af1738d74df827e45ba843a451fd3a25355527b9fd39b7feeeeed22ffa7fef9a

Created an InstanceManager to change Instances, but there are other ways to handle that. If it's part of the URL then it could be handled through the routes in index.tsx and switch via a call to `navigate()`, but that would require modifying all calls to `navigate()` throughout the entire app, unless there's some way to set a "base route" that's always prepended

Gtoasted:
Yeah, I wasn't sure if one monolitihc pr  or multiple fine-grained pr's would be better. I'll see if can merge your changes and #997 into this pr later.

You could just urlencode a colon if it comes to that, but I think some other form of configuration discovery (like a well-known uri) would be a better solution either way.

As far as I know, navigate does actually support a base path, but it can only be supplied by the router component, so it would be static and therefore useless. If the url-based approach sticks, it would indeed require modifying all calls to navigate and a, but such is life. Maybe I could make wrapper tags for those, but I'm gonna hold out on such an endeavour until I know that it might actually get merged (or I get bored and do it anyway)).

Gtoasted:
I don't really have the time to work on this further, and #999 looks promising, so I'll close this now.

Pecacheu:
That's fine, your contributions were very helpful in laying the groundwork for the multi instance features

PR #1003, opened 2026-03-07
feat: Accept autocomplete suggestion w/ tab

Pecacheu:
Based on suggestion from https://github.com/orgs/stoatchat/discussions/1418

- Pressing Tab or Enter accepts the current autocomplete suggestion
- If the autocomplete is an emoji, a second `:` press works too, similar to Discord

### DIFF
--- a/packages/client/components/ui/components/features/texteditor/codeMirrorAutoComplete.ts
+++ b/packages/client/components/ui/components/features/texteditor/codeMirrorAutoComplete.ts
@@ -1,6 +1,6 @@
 import {
+  acceptCompletion,
   autocompletion,
-  closeCompletion,
   currentCompletions,
 } from "@codemirror/autocomplete";
 import { EditorView, keymap } from "@codemirror/view";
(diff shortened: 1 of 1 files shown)


packages/client/components/ui/components/features/texteditor/codeMirrorAutoComplete.ts
```diff
@@ -137,14 +137,20 @@ export function codeMirrorAutoComplete(
     ],
   });
 
+  function accept(view: EditorView, key: string) {
+    const completions = currentCompletions(view.state);
+    if (
+      key === ":" ? completions[0]?.type === "emoji" : completions.length > 0
```
chrishultin, line 143: Let's not limit this to just the Emoji - we should keep the autocomplete tab-to-select behavior consistent.

Pecacheu: Well how this code works is basically, if the key is a `:`, then it only accepts emojis, but otherwise (aka if the key is tab because that's the only other time this function is called) it works as long as there are any completions to accept. It's using a ternary operator.

I figure that the `:` key only makes sense for emojis, but the tab key should work on all types of autocomplete.

chrishultin: Ah, good point.

chrishultin (approved):

PR #1013, opened 2026-03-10
feat: Add invite_only support, add invite code field

### LINKED ISSUE #639: Client doesn't support invite-only mode

Dadadah:
Adds support for invite_only. Reads the invite_only configuration field from the configuration object. Supports sending an invite code in the url like `/login/create/invite_code_here`. This functionality rhymes with existing functionality in other systems.

This PR also adds the ability to pass default values and disabling login form elements. Might be useful in some other cases.

Closes #639

Supersedes #769
Supersedes #957

### DIFF
--- a/packages/client/components/auth/src/flows/FlowCreate.tsx
+++ b/packages/client/components/auth/src/flows/FlowCreate.tsx
@@ -1,13 +1,13 @@
 import { Trans } from "@lingui-solid/solid/macro";
 
+import { useApi, useClient } from "@revolt/client";
 import { CONFIGURATION } from "@revolt/common";
-import { useNavigate } from "@revolt/routing";
+import { useNavigate, useParams } from "@revolt/routing";
 import { Button, Row, iconSize } from "@revolt/ui";
 
 import MdArrowBack from "@material-design-icons/svg/filled/arrow_back.svg?component-solid";
 
-import { useApi } from "../../../client";
-
+import { Show } from "solid-js";
 import { FlowTitle } from "./Flow";
 import { setFlowCheckEmail } from "./FlowCheck";
 import { Fields, Form } from "./Form";
@@ -17,7 +17,9 @@ import { Fields, Form } from "./Form";
  */
 export default function FlowCreate() {
   const api = useApi();
+  const getClient = useClient();
   const navigate = useNavigate();
+  const { code } = useParams();
 
   /**
    * Create an account
@@ -27,11 +29,13 @@ export default function FlowCreate() {
     const email = data.get("email") as string;
     const password = data.get("password") as string;
     const captcha = data.get("captcha") as string;
+    const invite = data.get("invite") as string;
 
     await api.post("/auth/account/create", {
       email,
       password,
       captcha,
+      ...(invite ? { invite } : {}),
     });
 
     // FIXME: should tell client if email was sent
@@ -43,6 +47,14 @@ export default function FlowCreate() {
     navigate("/login/check", { replace: true });
   }
 
+  const isInviteOnly = () => {
+    const client = getClient();
+    if (client.configured()) {
+      return client.configuration?.features.invite_only;
+    }
+    return false;
+  };
+
   return (
     <>
       <FlowTitle subtitle={<Trans>Create an account</Trans>} emoji="wave">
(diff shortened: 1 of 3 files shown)


Kirskov:
Hi,

I dont know why you want an UI field for the code when passing it through the http body is enough?

chrishultin (approved):

Dadadah:
Sorry I forgot to include pictures!

image

image

Here's inserting an invite into the URL:

image
image

calvinhg:
Would be cool if the invite code could be appended to the signup link too (`<domain.url>/login/create?invite_code={code}`)
edit: didn't see the last few screenshots! I see this is already in it

Kirskov:
> Would be cool if the invite code could be appended to the signup link too (`<domain.url>/login/create?invite_code={code}`)

That what I proposed in my PR above but they ignore my questionning.

So i'm done contributing.

Dadadah:
> Would be cool if the invite code could be appended to the signup link too (`<domain.url>/login/create?invite_code={code}`)

There are currently no other locations in the for-web repo that have this functionality so I am avoiding implementing it. Instead, the functionality rhymes with server invite and password reset functionality, which uses the url to retreive the code.



> Hi,
> 
> I dont know why you want an UI field for the code when passing it through the http body is enough?

The user should be able to see the invite code. Putting it in the http body only makes the code invisible to the user. If the user receives an error that the invite code is incorrect, they may not realize they're even using an invite code if it is not visible to the user.


> 
> That what I proposed in my PR above but they ignore my questionning.
> 
> So i'm done contributing.

Day job takes priority, sorry.

Dadadah:
DCO is failing as I committed from my work machine during a break. I will squash and resign after work this evening.

Kirskov:
> > Would be cool if the invite code could be appended to the signup link too (`<domain.url>/login/create?invite_code={code}`)
> 
> There are currently no other locations in the for-web repo that have this functionality so I am avoiding implementing it. Instead, the functionality rhymes with server invite and password reset functionality, which uses the url to retreive the code.
> 
> > Hi,
> > I dont know why you want an UI field for the code when passing it through the http body is enough?
> 
> The user should be able to see the invite code. Putting it in the http body only makes the code invisible to the user. If the user receives an error that the invite code is incorrect, they may not realize they're even using an invite code if it is not visible to the user.
> 
> > That what I proposed in my PR above but they ignore my questionning.
> > So i'm done contributing.
> 
> Day job takes priority, sorry.

Thank you for your answer. But what I dont get is : my PR takes the invite query parameter and add the code to the http form post request directly. It works perfectly on at least 2 self hosted instances I know of.

No field needed, the user copy the link, they create the account right away the code do not need to be print in a dedicated form field. 😅

But if your solution is the one required well, at least the server admin won't scratch their head when they want to invite someone.

mihaicm93 (approved):

packages/client/components/auth/src/flows/FlowCreate.tsx
```diff
@@ -43,13 +47,28 @@ export default function FlowCreate() {
     navigate("/login/check", { replace: true });
   }
 
+  const isInviteOnly = () => {
+    const client = getClient();
+    if (client.configured()) {
+      return client.configuration?.features.invite_only;
+    }
```
mdevnull, line 54: Currently running into an issue that my instance is configured for invite only and registration fails with the appropriate error but I dont see the invite code field. 

I also do see the /api/ call and that includes an `invite_only: true` feature flag. Is it possible if the page is rendered when the configuration isnt loaded yet this fails without any error/warning?

chrishultin: What version of the UI are you running?

mdevnull: `ghcr.io/stoatchat/for-web:103d2bf` That should be https://github.com/stoatchat/for-web/commit/103d2bf2e6f40a892fca4d51a31e518e3c968ddd so 0.3.0

chrishultin: You'll need a newer build - see https://github.com/stoatchat/self-hosted/issues/202#issuecomment-4050503968

PR #1020, opened 2026-03-13
fix: Dark theme flashbangs

### LINKED ISSUE #606: Flashbang in Dark Appearance Mode
### What happened?

There is a really bad white transition that happens when I click between channels in a server while having a dark appearance. [Here's a recording. ](https://drive.proton.me/urls/6MD6MP89DG#lHkghCoH9q6v)

Pecacheu:
Possibly fixes #606 (hard to tell because it's hardware dependent and thus hard to diagnose, but it worked on my Linux systems)

### DIFF
--- a/packages/client/components/ui/styles.css
+++ b/packages/client/components/ui/styles.css
@@ -22,6 +22,7 @@ body {
   left: env(safe-area-inset-left);
   right: env(safe-area-inset-right);
   bottom: env(safe-area-inset-bottom);
+  background: #191919;
 }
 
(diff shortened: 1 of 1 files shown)


Pecacheu:
Bruh wtf how did the i18n nuke go off, this is straight off the main branch with no changes to any files that have translated strings

Pecacheu:
> lgtm, just drop the i18n folder :trollface:

Because when everyone has translations... no one does. - Syndrome, probably

PR #1022, opened 2026-03-13
fix: Spacing around emoji/link/markdown in reply preview

### LINKED ISSUE #1182: [UI][REPLIES] Syntax-provided text offset
### What happened?

When you reply to a message with syntax included, it offsets the replied-message preview above the replying message—in my case; upwards. It seems the replied-message preview doesn't account for syntax.

Image

mihaicm93:
Fix reply preview removing spaces around emoji/link/markdown by ~~replacing ASCII space to unicode char to preserve the gap~~ removing the `display: "flex"` on ReplyContent and Link elements thanks to @Pecacheu for pointing it out

image (1)

***

Fixes: <https://github.com/stoatchat/for-web/issues/1182>

- `main` 
  - \#1022 :point\_left:

### DIFF
--- a/packages/client/components/ui/components/features/messaging/elements/MessageReply.tsx
+++ b/packages/client/components/ui/components/features/messaging/elements/MessageReply.tsx
@@ -110,8 +110,11 @@ export function MessageReply(props: Props) {
     if (content.length > 128) {
       content = content.slice(0, 128) + "...";
     }
+    content = content.replace(/\n/g, " ");
 
-    return renderSimpleMarkdown(content.replace(/\n/g, " "));
+    content = content.replace(/ /g, "\u00A0");
+
+    return renderSimpleMarkdown(content);
   };
 
   return (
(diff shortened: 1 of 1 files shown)


Pecacheu:
Wow that's a funny coincidence, I was just about to make a PR for this.

There's actually an easier way to solve it than injecting special characters into the string. If you just remove `display: "flex"` from **ReplyContent** and **Link** elements in the same file, the spaces reappear like magic. It was just classic flexbox weirdness.

mihaicm93:
> Wow that's a funny coincidence, I was just about to make a PR for this.
> 
> There's actually an easier way to solve it than injecting special characters into the string. If you just remove `display: "flex"` from **ReplyContent** and **Link** elements in the same file, the spaces reappear like magic. It was just classic flexbox weirdness.

ow, I will update to that after testing, thanks for that, it is much cleaner solution then mine

Pecacheu (approved):

Dadadah (approved):

PR #1024, opened 2026-03-13
feat: PWA hint dialog for mobile devices

Pecacheu:
Split off from #835

A lot of users, including my testers so far, have _literally no idea_ what the heck a Progressive Web App is, and didn't even know they could install the web page like an app. Unlike on desktop (where Chromium browsers for instance show a little 'install' icon in the URL bar), mobile browsers (Android and iOS) don't really display any hint whatsoever on supported sites.

This dialog should pop up for the user after logging in- but **only once and only on mobile**, with a helpful tip. Once dismissed, it should not appear again.

### DIFF
--- a/packages/client/components/modal/modals/TryPWA.tsx
+++ b/packages/client/components/modal/modals/TryPWA.tsx
@@ -1,0 +1,49 @@
+import { Trans } from "@lingui-solid/solid/macro";
+
+import { Dialog, DialogProps, iconSize } from "@revolt/ui";
+
+import MdInfo from "@material-design-icons/svg/outlined/error.svg?component-solid";
+
+import { t } from "@lingui/core/macro";
+import { useState } from "@revolt/state";
+import { Modals } from "../types";
+
+export function TryPWAModal(props: DialogProps & Modals & { type: "try_pwa" }) {
+  const state = useState();
+
+  return (
+    <Dialog
+      icon={<MdInfo {...iconSize(24)} />}
+      show={props.show}
+      onClose={() => {
+        state.settings.setValue("pwa:shown", true);
+        props.onClose();
+      }}
+      title={<Trans>Did you know?</Trans>}
+      actions={[
+        { text: <Trans>Close</Trans> },
+        {
+          text: <Trans>Install</Trans>,
+          onClick: () => {
+            // @ts-expect-error PWA event not recognized
+            if (state.pwaPrompt) state.pwaPrompt.prompt();
+            else
+              alert(
+                t`Sorry, your device doesn't support auto-install. Check your browser's menu for the 'Add to Home Screen' option.`,
+              );
+          },
+        },
+      ]}
+    >
+      <Trans>
+        You can <b>install</b> this Web App on your devive for conveient access
+        from the Home Screen, just like a real app. It even hides the huge menu
+        bar!
+        <br />
+        <br />
+        To install, tap the button below or open your browser's menu and choose
+        <b> Add to Home Screen</b>.
+      </Trans>
+    </Dialog>
+  );
+}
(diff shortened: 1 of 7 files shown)


packages/client/components/modal/modals/TryPWA.tsx
```diff
@@ -0,0 +1,49 @@
...
+            if (state.pwaPrompt) state.pwaPrompt.prompt();
+            else
+              alert(
+                t`Sorry, your device doesn't support auto-install. Check your browser's menu for the 'Add to Home Screen' option.`,
+              );
+          },
+        },
+      ]}
+    >
+      <Trans>
```
Dadadah, line 38: Does it make sense to put html tags inside a translation? I would assume not.

Dadadah, line 38: I think this is too much text. Might it be better to just say something like "Would you like to install this app to your home screen?"

Dadadah (changes requested):
Does this need to be a modal? If it's a modal, we have to rely on finding out if the device is mobile, which is hard to do. I get that users don't know anything about PWAs, but maybe it's better to just put something in the settings screen for it instead? That way it's not invasive and instead allows the user to do it if they look in the settings.

packages/client/components/modal/modals/TryPWA.tsx
Pecacheu: I was worried about that too, but it seems like lingui is designed to handle that. If you look at the translation files (despite the CI bot's best attempts to mangle them horribly) you can see they have substitution marks in them the same as lingui adds when adding a translation string that has a variable string inside it (which is done in various parts of the app already.)

Pecacheu: Haha that's fair, it might be a bit too wordy, I may be doing a bit too much r/ExplainLikeImFive. I'll come up with some shorter text later.

Pecacheu:
> I get that users don't know anything about PWAs, but maybe it's better to just put something in the settings screen for it instead? That way it's not invasive and instead allows the user to do it if they look in the settings.

An option in settings might be an alternative option, but I'm not sure where would be a good place to put it. Maybe as a dismissible banner that appears at the top when you open settings. Otherwise if it's buried under a random menu like Appearance then it defeats the purpose because non-tech-savvy users will have a hard time finding it, but on the other hand if it's front and center with its own category in the sidebar then it'll be annoying to always be in your way if you don't want to install the PWA.

> Does this need to be a modal? If it's a modal, we have to rely on finding out if the device is mobile, which is hard to do.

As for the detection method, the User Agent approach I'm using is very reliable and successfully detects any remotely modern Android or iOS device running any browser (and it would have to be decently modern to support PWAs anyway), it also has the advantage of not going off if the user toggles the "Desktop mode" setting on in their mobile browser.

The reason it's often frowned upon in the web design world to use User Agent detection is that narrower scope feature detection is usually better. (Eg. need your UI to adapt to a touch screen? Detect the touch screen, not just that it's a mobile device and assume. Or need a phone-scaled UI? Then trigger it based on the screen size and/or aspect ratio (which is what I'm doing in #835 btw.) But in this case what we're trying to detect is simply that the device isn't a desktop PC, since mobile users are much more "app-brained."

Pecacheu:
Less wordy dialogue as per Dadadah.


Alternatively, we could go with a "No Thanks" button to make the one-time nature of the popup more clear.

Misofist:
I agree that this should go in the Settings menu. I don't think it should be a dismiss-able banner, nor should it be buried in a submenu. I think there should be an "Install" button in the settings menu, preferably right near the top of the list.

Tapping on this should immediately fire `prompt()` in browsers which support `beforeinstallprompt`. In browsers which do not support this, it can open be a settings page with instructions on how to install it manually.

Pecacheu:
Superseded by #1077

PR #1025, opened 2026-03-13
feat: Voice Call Button in header

### LINKED ISSUE #541: feat: hide/toggle "Join Call" card
It would be nice to hide the "Join the voice call" card because it's a bit annoying for some users, and would like a simple button on the channel's header instead.

Pecacheu:
_Split off from #835_

Replaces the obnoxious Voice Call preview card with a button in the header for DMs

But retains it in server voice channels (as requested by Insert)

Fixes #541


Pecacheu:
Annnnd of course there's merge conflicts in the strings even tho I didn't add any new ones. Why not?

Dissssy:
I love the stoat devs but i'm pretty sure this is like the third open PR for this specific ui feature in the last month lol

Pecacheu:
@Dissssy Technically it was the first, but they told me to split off unrelated things from my mobile UI mega-PR into their own PRs. I ended up fixing this early on as it was *way* more annoying on a small phone screen.

gabrielfordevelopment:
Can you provide a full screenshot?

Pecacheu:
@gabrielfordevelopment Sure, here you go

Out of call on server (unchanged)


Out of call on DM:


In call on DM:

gabrielfordevelopment:
Thank you so much for the provided media files.
It looks a bit weird in voice channels that there is a duplicated button that does the same thing.
image

Pecacheu:
> Thank you so much for the provided media files. It looks a bit weird in voice channels that there is a duplicated button that does the same thing.

I think it looks fine that way, but we could make it hide when in a call. Question for UI/UX

packages/client/src/interface/channels/ChannelHeader.tsx
```diff
@@ -134,6 +136,26 @@ export function ChannelHeader(props: Props) {
 
       <Spacer />
 
+      <Show when={props.channel.isVoice}>
```
Dadadah: This should not be visible if the call card is visible.

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
```diff
@@ -306,16 +306,23 @@ function VoiceCallCard(props: { channel: Channel }) {
...
-          fallback={<VoiceCallCardPreview channel={props.channel} />}
-        >
-          <VoiceCallCardActiveRoom />
-        </Show>
-      </Card>
-    </Base>
+    <Show
+      when={
+        inCall() ||
+        (props.channel.type === "TextChannel" && props.channel.isVoice)
```
Dadadah: Does `inCall` only return true if you are in the call?

If so, the call card should also show if any member of the dm or group is in the call. When the call card is shown, the call button should be removed.

The logic for this check should probably exist on the voice context since it will be used here and if the call button is visible.

Dadadah (changes requested):

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Pecacheu: That's a good idea, which *I think* could be considered in the scope of this PR. Granted I get yelled at a lot for putting too much in one PR lol

packages/client/src/interface/channels/ChannelHeader.tsx
Pecacheu: This is what @gabrielfordevelopment suggested, but as mentioned it is a subjective opinion and a question for UI/UX, eg. @infi

Dadadah: I disagree. There is no reason to display the call button on the top row if it exists in the call card.

infi:
Resolutions on disagreements above:
- The call card should always be visible whenever there is an ongoing call in a channel
- There must not be a second "end call" button on the screen, that feature is part of the extended call card

packages/client/src/interface/channels/ChannelHeader.tsx
Pecacheu: Fixed

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Pecacheu: Fixed

Dadadah (approved):

PR #1027, opened 2026-03-14
feat: Embeds allowed directly in messages to enable powerful Markdown formatting

### LINKED ISSUE #998: Skip this prompt

Pecacheu:
_Split off from #835_

The old style of placing all embeds at **end** of the message can lead to confusion, for instance the embeds are often out of order from the links themselves, such as the example below.

This alternative approach places the embeds directly into the message content like a more traditional Markdown style, allowing for much clearer communication, with relevant images/links placed right where they're needed! To prevent an embed from appearing, you can use a Markdown link with custom text.

This fails safe in both directions: Embeds that the client can't find a matching link for still appear at the end as normal, and links that don't have a matching embed appear as a normal anchor link.

Also fixes this erroneous warning when clicking embed links.

Fixes #998

### DIFF
--- a/packages/client/components/app/interface/channels/text/Message.tsx
+++ b/packages/client/components/app/interface/channels/text/Message.tsx
@@ -18,7 +18,6 @@ import { styled } from "styled-system/jsx";
 import { decodeTime } from "ulid";
 
 import { useClient } from "@revolt/client";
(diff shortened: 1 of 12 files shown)


insertish:
This seems like a logical continuation of the invite handling, lets see what UI/UX thinks haha

PR #1029, opened 2026-03-14
fix: Make dialogs appear above User Card & User Profile, modals consistently auto-dismiss User Card

### LINKED ISSUE #858: Edit identity modal covered up
### What happened?

When clicking on my picture in a channel, clicking the three dots and selecting "edit identity" the modal popup for editing my identity gets hidden behind the modal for my profile unless i click out of my profile. This only happens on smaller window sizes.

Steps to replicate:
1. Make sure window is small size (size of window in testing was 1/4 of my screen size 2560 x 1440 (1280 x 720) windows display scaling is set to 100% too.
2. Click on profile picture after sending a message in a channel.
3. Click the three dot hamburger menu button.
4. Click edit identity
5. Modal pops up behind profile modal.

Video of bug:

https://github.com/user-attachments/assets/181eec71-e035-4644-a216-9e2d941cc605

### LINKED ISSUE #1028: Edit Roles menu appears below User Card in hierarchy
### What happened?

Idk how no one reported this yet but I couldn't find an existing one from trying a few different search terms, so I'm opening a report and PR for this. If you open the *Edit Roles* menu from the *User Card*, the former appears *below* the latter, making it unusable without carefully closing the user card and trying not to close the roles menu at the same time in the process.

Pecacheu:
Fixes #1028

Before & After
 

Added by maintainers:
Fixes #858

### DIFF
--- a/packages/client/components/ui/components/floating/FloatingManager.tsx
+++ b/packages/client/components/ui/components/floating/FloatingManager.tsx
@@ -174,7 +174,7 @@ function Floating(props: FloatingElement & { mouseX: number; mouseY: number }) {
           position: position.strategy,
           top: `${position.y ?? 0}px`,
           left: `${position.x ?? 0}px`,
-          "z-index": (999 - (props.show()?.userCard ? 1 : 0)).toString(),
+          "z-index": props.show()?.userCard ? "99" : "999",
         }}
       >
         <Switch>
--- a/packages/client/components/ui/components/floating/FloatingManager.tsx
+++ b/packages/client/components/ui/components/floating/FloatingManager.tsx
@@ -174,7 +174,7 @@ function Floating(props: FloatingElement & { mouseX: number; mouseY: number }) {
           position: position.strategy,
           top: `${position.y ?? 0}px`,
           left: `${position.x ?? 0}px`,
-          "z-index": props.show()?.userCard ? "998" : "999",
+          "z-index": props.show()?.userCard ? "99" : "999",
         }}
       >
         <Switch>
--- a/packages/client/components/ui/components/floating/FloatingManager.tsx
+++ b/packages/client/components/ui/components/floating/FloatingManager.tsx
@@ -174,7 +174,7 @@ function Floating(props: FloatingElement & { mouseX: number; mouseY: number }) {
           position: position.strategy,
           top: `${position.y ?? 0}px`,
           left: `${position.x ?? 0}px`,
-          "z-index": props.show()?.userCard ? "99" : "999",
+          "z-index": props.show()?.userCard ? "998" : "999",
         }}
       >
         <Switch>
--- a/packages/client/components/modal/index.tsx
+++ b/packages/client/components/modal/index.tsx
@@ -11,6 +11,7 @@ import { SetStoreFunction, createStore } from "solid-js/store";
 import type { MFA, MFATicket } from "stoat.js";
 
 import { Keybind, KeybindAction } from "@revolt/keybinds";
+import { dismissFloatingElements } from "@revolt/ui";
 
 import { RenderModal } from "./modals";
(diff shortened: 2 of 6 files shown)


gabrielfordevelopment:
I checked it out and it is inconsistent in the current system, since you have two options to go to the "give role" modal:

(You didn't touched this, just saying that it exist)
Click on profile → click on "Roles" - then it opens up the role editor and closes all other components. 
image

(This is what you changed)
However, if you do Profile → 3 dots → Edit Roles it does NOT close the other components
image


This is inconsistent. Consider: close all components here too instead of adjusting the Z index to be in front.

Pecacheu:
> This is inconsistent. Consider: close all components here too instead of adjusting the Z index to be in front.

Ah, I didn't even realize you could click that to edit roles lol

You're right, it's two different behaviors when both options should act in the same way. I _can_ change it to dismiss the User Card instead of alter the z-index, but now, since both behaviors exist in the current build, the question becomes... which one is the **intended** way to handle it from a UI perspective... Roles open as a separate sub-menu, or roles open as a new overlay that replaces the old one?

Probably a question for the UI/UX team.

Dadadah (changes requested):
Doesn't it make sense to have all floating elements below the dialog? The dialog should always be the most up front thing.

I think this line of code originally made it where other floating elements (like context menus) would show up on top of the user profile menu. However, dialogs actually set their own z-index in `packages/client/components/ui/components/design/Dialog.tsx:128` and I think it makes more sense that this 100 should actually be 1000 so it will appear above all floating elements. In fact, I would argue that was the original intention and the fact that it is 100 instead of 1000 is a typo.

Pecacheu:
> Doesn't it make sense to have all floating elements below the dialog? The dialog should always be the most up front thing.

I agree with this for sure. I think our above question for UI is more about whether the user card should auto-close once you hit edit roles or it should stay open in the background- But in either case, the dialog should have a higher z-index and render above it. I can change that zIndex to 1000 just to make sure it does.

packages/client/components/ui/components/floating/FloatingManager.tsx
```diff
@@ -174,7 +174,7 @@ function Floating(props: FloatingElement & { mouseX: number; mouseY: number }) {
           position: position.strategy,
           top: `${position.y ?? 0}px`,
           left: `${position.x ?? 0}px`,
-          "z-index": (999 - (props.show()?.userCard ? 1 : 0)).toString(),
+          "z-index": props.show()?.userCard ? "99" : "999",
```
Dadadah, line 177: I think you can just get rid of this change all together tbh

Dadadah:
> I agree with this for sure. I think our above question for UI is more about whether the user card should auto-close once you hit edit roles or it should stay open in the background . . .

If you click on the role modal, it dismisses the profile window. Therefore I'm going to call this a bug, and say that *yes* the floating elements should be dismissed (calling the `dismissFloatingElements()` like `packages/client/components/ui/components/features/profiles/ProfileRoles.tsx:28`) and merge this without ui/ux as it's not necessary for bug fixes.

packages/client/components/ui/components/floating/FloatingManager.tsx
Pecacheu: Ok I can put it back to 999 and 998

Pecacheu:
Welp, here I went on this whole complicated thing with ensuring that the full user profile (which uses the same actions menu) doesn't get auto-dismissed as before, but the user card does, only to realize that `dismissFloatingElements()` doesn't dismiss the user profile and only the card, so we're already good I can just drop those if statements I added lol

Pecacheu:
Kay fixed, the simple way now

Dadadah (approved):

PR #1031, opened 2026-03-14
fix: Make the sync store only update if remote is different than local

### LINKED ISSUE #1009: Triggering UserSettingsUpdate event causes infinite loop
### What happened?

Doing anything that causes the user settings to update (eg. reordering servers) triggers an infinite loop of requests to the /api/sync/settings/set endpoint only stopping if eventually the endpoint returns an error (usually 429, but on the main stoat.chat instance it seems to go on forever)

Presumably updating your settings causes the backend to emit a UserSettingsUpdate event which trips your client to update its client side copy of the settings, but it's doing it in a way that re-emits the request? that's my best guess.

Dadadah:
Adds a deep equal function to Ordering and NotificationOptions stores. The sync store will not attempt to update the server if the incoming data is equal to the existing data.

Closes #1009

### DIFF
--- a/packages/client/components/state/stores/NotificationOptions.ts
+++ b/packages/client/components/state/stores/NotificationOptions.ts
@@ -86,6 +86,15 @@ export class NotificationOptions extends AbstractStore<
   }
 
   /**
+   * Get this store's value
+   *
+   * Reexported to allow equals checking for syncing
+   */
+  get() {
+    return super.get();
+  }
+
+  /**
    * Hydrate external context
    */
   hydrate(): void {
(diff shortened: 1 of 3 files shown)


chrishultin (approved):

mihaicm93 (approved):

insertish (commented):
Maybe instead of writing out our own comparison functions, we just use lodash isEqual?

Dadadah:
I'm not sure we can use isEqual when using states like this. The state doesn't return a pure object so it won't be equal with the parsed object. Is there a way to get the pure object out of the state?

Edit: Alternatively, should I write a method to get the pure object out of the state for comparison with isEquals?

Dadadah:
I've refactored this to use lodash isequals by having the notification and ordering stores export the `get` function to allow getting the raw object directly from the store.

PR #1032, opened 2026-03-16
chore: properly handle bulk deletion in message view

amycatgirl:


packages/client/components/app/interface/channels/text/Messages.tsx
```diff
@@ -671,17 +671,34 @@ export function Messages(props: Props) {
...
+      );
+    }
+  }
 
   // Add listener for messages
   onMount(() => {
     const c = client();
     c.addListener("messageCreate", onMessage);
     c.addListener("messageDelete", onMessageDelete);
+    c.addListener("messageDeleteBulk", onMessageDeleteBulk);
```
chrishultin, line 695: Am I missing something here?  Does this ever get called?

amycatgirl: It does, but only when the element is first mounted. The `onMessageDeleteBulk` function is only called when a bulk deletion event is received by the client. The listener is also removed when the element is unmounted.

It worked on my (very limited) testing, used Automod to bulk delete 10 messages and could confirm that the message elements did not persist.

Pecacheu:
I wonder if this also solves the issue of messages deleted by the Bolt bot not actually disappearing on the Stoat end until refresh... But I would assume that's a regular delete API call and not a bulk delete one?

Also, don't forget to drop the i18n files from this one now that we have that workflow fixed.

amycatgirl:
i forgor :skull:

amycatgirl:
I would need to check what the bolt bot does when it deletes messages to see if this is the same issue.

amycatgirl:
Looks like that's the case with bolt: https://codeberg.org/jersey/lightning/src/commit/6a1e2a2b815f20842fa6ba78c9a56882d33298bb/pkg/platforms/stoat/plugin.go#L177-L184

amycatgirl:
Looks like this is still the case on 0.6.0, just tested the branch again and it still works.

Pecacheu:
Worked well when I tested it on a previous version too, I forgot to approve it

Pecacheu (approved):

PR #1037, opened 2026-03-17
fix: properly check if link is in scope of app

amycatgirl:
supersedes #737 because i completely fucked up the branch lol

also takes into account the review inside of the previous PR


mihaicm93 (approved):

chrishultin (approved):

Pecacheu:
Not sure if a hard-coded list of endpoints like that is the *ideal* way to detect it, since it will have to be updated if we add any more in the future, but it'll definitely do for now to solve the issue.

Pecacheu:
Also if/when this gets merged I will integrate it into #999 because checking the scope of links is important there. Though in that PR it will require a small change of adding the URL of the current instance (if not Stoat) to the list. Currently that is not necessary because the instance can be assumed to be Stoat unless someone rebuilds the client.

Dadadah:
@amycatgirl can you remove the translations from this pr? The workflow has since been fixed.

amycatgirl:
I removed the translation commit. I am now going to rebase the branch into the latest commit

PR #1047, opened 2026-03-21
ci: workflow clean up and fixes

insertish:
- ci: only check lingui in build & test ci: generate catalogs in release please workflow
- ci: only run documentation workflow on main
- ci: only run git town on non-fork PRs
- ci: un-used branch trigger

PR #1048, opened 2026-03-21
fix: Github workflows & rogue i18n bot

Pecacheu:
Changes:
- Lingui/i18n workflow is now in it's own file and only triggers on push to main branch
- Build documentation only triggers on push to main branch
- Removed display branch stack because it seems to fail most of the time (If there's another way to fix it lmk)

Imo we should perhaps even do the _opposite_ of what the lingui bot was set up to do before: Make it reject PRs that change the translation files (before merge that is) unless they are specifically a translation PR. That would help avoid accidentally breaking the translation files. Which can _absolutely happen_ from i18n build running out of control on everyone's branches lol.


Pecacheu:
Superseded by #1047

PR #1059, opened 2026-03-24
fix: Strip HTML from error strings

Pecacheu:
Now ideally the Stoat backend should just not send API errors in HTML format to begin with (but as raw text or JSON, for instance) because it wastes network bandwidth, but hey, stripping it on the frontend also helps in case any HTML-formatted errors from other sources slip through the cracks. Which- if other people's error report screenshots are anything to go by- happens a lot.

Old:

New:

### DIFF
--- a/packages/client/components/i18n/errors.ts
+++ b/packages/client/components/i18n/errors.ts
@@ -1,6 +1,8 @@
 import { useLingui } from "@lingui-solid/solid/macro";
 import { API } from "stoat.js";
 
+const RE_BREAK = /\s*\n\s*/g;
+
 /**
  * Translate any error
  */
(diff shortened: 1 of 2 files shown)


Zomatree:
If the client adds `Accept: application/json` the api will return json errors for all errors.

Pecacheu:
> If the client adds `Accept: application/json` the api will return json errors for all errors.

Ah, good to know. I assume that's a change for stoat.js and not for-web though?

chrishultin:
> > If the client adds `Accept: application/json` the api will return json errors for all errors.
> 
> Ah, good to know. I assume that's a change for stoat.js and not for-web though?

Yeah, that'd probably be the place - but making sure we're not throwing it up to the user in the UI is probably a good idea.

chrishultin:
https://github.com/stoatchat/javascript-client-sdk/issues/165 for the longer-term fix

chrishultin (approved):

Pecacheu:
Ok don't think we need UI/UX approval for this one since it's not a UI change to so I'll merge it

PR #1061, opened 2026-03-25
fix: incorrect scaling (and repeated disconnects) for live video tracks taller than 16:9

Misofist:
The existing CSS uses object-fit: contain; with both width & height set to 100% in an attempt to scale video tracks to fit within the intended space. But if the video element is too tall, then it was simply causing the <video> element to overflow.

This results in only the top 16:9 rectangle of the video being visible. Additionally, if the visible area is <80% of the video, then the visibility observer causes the video track to repeatedly be unsubscribed / resubscribed.

By my maths, this affected aspect ratios taller than 64:45. Most notably, square videos, and all vertical videos.

(This one-liner fix took me hours to track down 😭)

## To reproduce

Place two users in a call. Have one user stream a vertical video (you can do this using a webcam in portrait mode, or by screen-sharing an application that's taller than it is wide). On the other user's end, you should be able to see the bug clearly.

### DIFF
--- a/packages/client/components/app/interface/settings/channel/Overview.tsx
+++ b/packages/client/components/app/interface/settings/channel/Overview.tsx
@@ -92,6 +92,9 @@ export default function ChannelOverview(props: ChannelSettingsProps) {
           </Text>
           <Form2.FileInput control={editGroup.controls.icon} accept="image/*" />
           <Form2.TextField
+            minlength={1}
+            maxlength={32}
+            counter
             name="name"
             control={editGroup.controls.name}
             label={t`Channel Name`}
@@ -99,6 +102,8 @@ export default function ChannelOverview(props: ChannelSettingsProps) {
           <Form2.TextField
             autosize
             min-rows={2}
+            maxlength={1024}
+            counter
             name="description"
             control={editGroup.controls.description}
             label={t`Channel Description`}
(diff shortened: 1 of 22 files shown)


Misofist:
Didn't mean to close this PR, was just having an issue with my Forgejo <-> GitHub mirroring.

Dadadah:
This change works great on my self host, can you please add it to the screen share styling as well? It's located near line 258 in the same file.

Misofist:
Fixed the linter error, should be ready to go now. :)

Dadadah:
This was merged in #1090. Closing. Thanks for the help @Misofist

Misofist:
I'm glad the fix landed, but it's also a bit of a shame that the squash removes me from the commit history, and that this PR shows as closed without merging, and that Release Please won't include this fix in release notes.

Not a big deal to me, but I just wanted to raise this as something that might be worth considering in future, especially for first-time contributors.

Dadadah:
I tried to get you in to it, I don't fully understand how Github decides who gets shown. I figured if I tagged you it'd be enough but it wasn't. I'll figure it out next time because I understand the feeling.

mihaicm93:
We can just edit the release please and add this PR manually to it

Misofist:
> I'll figure it out next time because I understand the feeling.

The best way is to merge each individual PR individually, instead of merging them into each other and then merging one of them into the main branch.

This ensures that the commit history contains each PR author, the PRs on GitHub are marked as Merged instead of just Closed, the PR author appears on the Contributors tab (https://github.com/stoatchat/for-web/graphs/contributors), Release Please will include the change, etc...



Alternatively, even if a PR was merged via another PR, project maintainers can still merge the other PRs. It'll show a merge conflict (because the changes already exist) but you can proceed with merging the branch, even if the merge commit ends up containing no changes. This will still do all of the above things I mentioned. :)

PR #1062, opened 2026-03-25
fix: unify autocomplete scrollbar styling via shared scrollable classes

gabrielfordevelopment:
Applied the app’s shared scrollbar styling to CodeMirror autocomplete tooltips for consistent and maintainable behavior.

`scrollable.ts`  - exported shared scrollableStyles for reuse  
`codeMirrorAutoComplete.ts`  - applied shared scrollable classes to autocomplete tooltip classes (emoji, mention, role, channel...)

---

## Currently we have this:

chrome_O16XiFOkvv

---

## What I made have this:

chrome_a6C6Lo4Bzj

### DIFF
--- a/packages/client/components/ui/components/features/texteditor/codeMirrorAutoComplete.ts
+++ b/packages/client/components/ui/components/features/texteditor/codeMirrorAutoComplete.ts
@@ -5,6 +5,8 @@ import {
 } from "@codemirror/autocomplete";
 import { EditorView, keymap } from "@codemirror/view";
 import { Accessor } from "solid-js";
+
+import { scrollableStyles } from "../../../directives/scrollable";
 import { AutoCompleteSearchSpace } from "../../utils/autoComplete";
 import { codeMirrorAutoCompleteSource } from "./codeMirrorAutoCompleteSource";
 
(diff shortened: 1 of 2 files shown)


mihaicm93 (approved):

gabrielfordevelopment:
> This only fixes the scrollbar for the emojis, it should fix the scrollbar on all of the autocompletes.


Thanks for the feedback, great point about reusing the shared scrollbar styling.

I updated the implementation to use the shared scrollable style classes instead of exporting a standalone color token. The autocomplete tooltip classes now include the shared scrollable classes for all completion types (emoji, mention, role, channel), so the same scrollbar styling is consistently applied across the CodeMirror autocomplete variants.

(Also, even in the previous version, the scrollbar change was applied via the shared autocomplete list selector, so mentions and roles were already affected.)

chrome_jWSD7AuOzr

Dadadah (approved):
LGTM! You could maybe remove the SCROLLBAR_COLOR const since it's not exported anymore but I won't block on that.

gabrielfordevelopment:
> LGTM! You could maybe remove the SCROLLBAR_COLOR const since it's not exported anymore but I won't block on that.

Done! Thank you so much for your patience!

chrishultin (approved):

PR #1071, opened 2026-03-28
fix: client.configured reference in signup flow

bundle-o-cells:
Broken reference caused an error when clicking sign up.

### DIFF
--- a/packages/client/components/auth/src/flows/FlowCreate.tsx
+++ b/packages/client/components/auth/src/flows/FlowCreate.tsx
@@ -49,7 +49,7 @@ export default function FlowCreate() {
 
   const isInviteOnly = () => {
     const client = getClient();
-    if (client.configured()) {
+    if (client.configured) {
       return client.configuration?.features.invite_only;
     }
     return false;


packages/client/components/auth/src/flows/FlowCreate.tsx
```diff
@@ -49,7 +49,7 @@ export default function FlowCreate() {
 
   const isInviteOnly = () => {
     const client = getClient();
-    if (client.configured()) {
```
amycatgirl, line 52: This seems wrong.

`client.configured` is a function (accessor) that returns whether the client properly fetched the current instance's configuration from the root endpoint. Functions, when not called, are always truthy values, since they are objects (see [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) on MDN)

So checking against it would useless.

I think using a [derived signal](https://docs.solidjs.com/concepts/derived-values/derived-signals) would be better in this case, since it relies on that `configured` signal.

bundle-o-cells: yep, you're right. The stoat.js submodule was out of date on my end and I misunderstood the error I received. Sorry for the bother.

PR #1072, opened 2026-03-28
feat(auth): skip email check flow when verification is disabled

Slx-hub:
Hey i tumbled in here from the Self-Host community! The Readme stated that with email verification disabled the Page still prompts the user to check his mail, a message that i have to forward to every user during onboarding. So i set out to fix that!

It simply checks the client config, analog to "isInviteOnly", and then either routes the old way or directly executes login for the user with the readily available credentials and then forwards him to the username selection.

I tested it by adding and removing an [api.smtp] connection to the toml for a local mailpit instance.

UI flow without email verification:
stoat_no_mail

UI flow with email verification:
stoat_with_mail

### DIFF
--- a/packages/client/components/auth/src/flows/FlowCreate.tsx
+++ b/packages/client/components/auth/src/flows/FlowCreate.tsx
@@ -1,9 +1,10 @@
 import { Trans } from "@lingui-solid/solid/macro";
 
-import { useApi, useClient } from "@revolt/client";
+import { useApi, useClient, useClientLifecycle } from "@revolt/client";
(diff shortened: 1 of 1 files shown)


Dadadah:
I think this is a duplicate of #760

Slx-hub:
> I think this is a duplicate of #760

Ah well, it sure is! 😅
I take it as learning experience..
I'd argue that mine looks a little nicer but functionally it's exactly the same. so as long as it's getting fixed

Dadadah:
Thanks for your contribution. This was fixed in the merging of #760 .

PR #1076, opened 2026-03-29
feat: Add create channel/category opts to respective context menus

Pecacheu:
Adds the "create channel" button to the category context menu and the "create category" to the channel context menu.

This makes it easier to add new ones on small screens or when the the channel sidebar is full, because there's no "blank space" to right-click on. I also made it so that adding a channel by right-clicking a category adds the new channel to _that_ category instead of the default at the top, since that makes more intuitive sense.

Before:

After:


chrishultin (approved):

infi (approved):
Finally!

Dadadah (approved):

PR #1077, opened 2026-03-29
feat: Guide users to install PWA

Misofist:
This builds on the work done by Pecacheu in #1024, but completely reworks the UX.

NOTE: A lot of the screenshots below look ugly, that's because this is designed to work alongside #835. It looks good when that PR is also included, I just didn't want to misrepresent what's included within this PR in my screenshots.

No modal window is shown. Instead, a new "Install" option has been added to the Settings menu.

On browsers that support the `beforeinstallprompt` event, this page contains a button to install the PWA.

Screenshot_20260330-054113

Clicking this button starts the PWA install experience:

Screenshot_20260330-054119

This also gracefully handles the case where the user declines the installation:

Screenshot_20260330-055635

For browsers which do not support the `beforeinstallprompt` event, manual installation instructions are shown:

image

The install button is hidden within the PWA itself:

image

The install button also only appears on mobile browsers, since currently on desktop I think there is little point in installing the PWA.

I also updated the way we use Vite PWA to follow their best practices for SW updates: https://vite-pwa-org.netlify.app/guide/periodic-sw-updates.html#handling-edge-cases


DeclanChidlow:
For the 'refresh to try again', a refresh button could be made present on the page for ease of use.

Also, I wrote a full article on detecting and styling based on if the app is installed as a PWA which could well be of use: https://www.smashingmagazine.com/2025/08/optimizing-pwas-different-display-modes/

Misofist:
> For the 'refresh to try again', a refresh button could be made present on the page for ease of use.

Good idea! I'll add that in when I'm at my computer. 

> Also, I wrote a full article on detecting and styling based on if the app is installed as a PWA which could well be of use: https://www.smashingmagazine.com/2025/08/optimizing-pwas-different-display-modes/

Great article. I don't *think* this changes anything, since Stoat's formatting is handled via JS and not via stylesheets. This is also intended to pair with #835 or similar, so I expect that the PWA manifest will only specify `standalone` mode.

But let me know if I've missed anything. :)

DeclanChidlow:
Ah, GitHub had a fart and didn't show me the full file, only the setting for iOS bit so I thought it'd been missed. Disregard.

Pecacheu:
This does seem like a better way of doing it yea, I'll test it out soon

Pecacheu:
Still need to test this one and looks like there's some conflicts to fix and you need to drop the messages.po files, but lemme mark the other one closed.
Supersedes #1024

Misofist:
I've just moved house so I'll be busy for a little while getting that all sorted. I'll revisit this to make the requested changes when I can.

Pecacheu:
FYI I finally gave this a proper test, there was a minor issue where the events didn't always fire, seemingly because the module-level indext.tsx script doesn't always load fast enough. So I realized that the events need to be hooked from an inline script.

I was going to make a change request but then I realized that much more substantial changes were required to get this rebased on main since there's been so many commits since. So I went ahead and did the rebase on this branch here: https://github.com/stoatchat/for-web/tree/pwainstall

Pecacheu:
Superseded by #1380

PR #1084, opened 2026-04-02
fix: remove scrollbarGutter stable from showOnHover to prevent carets without overflow

wiliyam:
## Summary

The member list sidebar showed scrollbar carets on hover even when there weren't enough members to require scrolling.

## Root cause
`scrollableStyles` applied `scrollbarGutter: 'stable'` when `showOnHover` was enabled. The `stable` value permanently reserves space for the scrollbar track — including its carets — regardless of whether content overflows.

## Fix
Removed `scrollbarGutter: 'stable'` from the `showOnHover` variant. The hover handlers already apply `overflow: scroll` on mouse entry, so scrolling still works correctly when content does overflow. The carets no longer appear when there's nothing to scroll.

### DIFF
--- a/packages/client/components/ui/directives/scrollable.ts
+++ b/packages/client/components/ui/directives/scrollable.ts
@@ -83,10 +83,17 @@ export function scrollable(
     const showClass = hoverStyles({ direction: props.direction }).split(" ");
 
     /**
-     * Handle mouse entry
+     * Handle mouse entry — only show scrollbar if content actually overflows.
+     * This prevents scrollbar carets from appearing when there's nothing to scroll.
      */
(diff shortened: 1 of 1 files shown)


wiliyam:
Good catch @Dadadah! The previous fix removed `scrollbarGutter: stable` but the hover handler still applied `overflow: scroll` unconditionally, which caused the browser to render the scrollbar track (and carets) even with no overflow.

Fixed — the `onMouseEnter` handler now checks `scrollHeight > clientHeight` (or `scrollWidth > clientWidth` for horizontal) before adding the scroll class. Carets will only appear when there's actually content to scroll.

PR #1085, opened 2026-04-02
refactor: Dedicated component for category select-one dropdowns

Pecacheu:
This makes the code much cleaner and probably a tad more performant by use of a dedicated component for select-one collapses instead of using a regular collapse and manually coding the select-one functionality separately _in every instance._

Instead of building out the individual child CategoryButtons, this one takes in an object of options and allows you to get/set the current value.

I switched from checkboxes to radios, fixed a bug where ticking the checkbox instead of clicking the body of the element causes weird behavior instead of selecting it, and changed the color scheme of inner elements to make it clear where the select actually starts and ends. I used the material tertiary color options for this, because it's really the only theme color that's not taken elsewhere.

Before:

After:

Reacts to current theme:

This change was applied to the voice and language settings tabs. I also replaced `CategoryButtonGroup` with `CategoryButton.Group` in a few files because it was already marked as depricated.

### DIFF
--- a/packages/client/components/app/interface/settings/user/Language.tsx
+++ b/packages/client/components/app/interface/settings/user/Language.tsx
@@ -6,8 +6,8 @@ import { timeLocale } from "@revolt/i18n/dayjs";
 import { UnicodeEmoji } from "@revolt/markdown/emoji";
 import { useState } from "@revolt/state";
 import {
-  CatSelOption,
   CategoryButton,
+  CategorySelectOption,
   Column,
   Row,
   Time,
@@ -52,7 +52,7 @@ function PickLanguage() {
   const { i18n } = useLingui();
 
   //@ts-expect-error unfilled object
-  const langOpts: { [k in Language]: CatSelOption } = {};
+  const langOpts: { [k in Language]: CategorySelectOption } = {};
   const langIds = Object.keys(Languages) as Language[];
 
   //Move user's system language to top
--- a/packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
+++ b/packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
@@ -4,7 +4,13 @@ import { useMediaDeviceSelect } from "solid-livekit-components";
 import { Trans } from "@lingui-solid/solid/macro";
 
 import { useState } from "@revolt/state";
-import { CategoryButton, CatSelOption, Column, Slider, Text } from "@revolt/ui";
+import {
+  CategoryButton,
+  CategorySelectOption,
+  Column,
+  Slider,
+  Text,
+} from "@revolt/ui";
 import { Symbol } from "@revolt/ui/components/utils/Symbol";
 
 /**
@@ -56,7 +62,7 @@ function SelectInput(props: { kind: MediaDeviceKind }) {
 
   const devOpts = createMemo(() => {
     const devs = media().devices(),
-      opts: { [k in string]: CatSelOption } = {};
+      opts: { [k in string]: CategorySelectOption } = {};
 
     //Ensure default is at top
     let d = devs.find((d) => d.deviceId === "default");
--- a/packages/client/components/ui/components/design/CategoryButton.tsx
+++ b/packages/client/components/ui/components/design/CategoryButton.tsx
@@ -346,7 +346,7 @@ CategoryButton.Collapse = (props: CollapseProps) => {
   );
 };
 
-export type CatSelOption = Omit<
+export type CategorySelectOption = Omit<
   ComponentProps<typeof CategoryButton>,
   "onClick" | "children"
 > &
@@ -366,7 +366,7 @@ type SelectProps<T extends string> = Omit<
   "onClick" | "children" | "description"
 > & {
   title?: JSX.Element;
-  options: { [k in T]: CatSelOption };
+  options: { [k in T]: CategorySelectOption };
   value?: T;
   onUpdate: (v: T) => void;
 };
--- a/packages/client/components/ui/components/design/index.ts
+++ b/packages/client/components/ui/components/design/index.ts
@@ -31,7 +31,7 @@
 export { Avatar } from "./Avatar";
 export { Badge } from "./Badge";
 export { Button } from "./Button";
-export { type CatSelOption, CategoryButton } from "./CategoryButton";
+export { type CategorySelectOption, CategoryButton } from "./CategoryButton";
 export { Checkbox } from "./Checkbox";
 export { DataTable } from "./DataTable";
 export { type DialogProps, Dialog } from "./Dialog";


packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
```diff
@@ -18,101 +14,73 @@ export function VoiceInputOptions() {
...
 
 /**
- * Select audio input
+ * Select input device w/ type
  */
-function SelectMicrophone() {
-  const { t } = useLingui();
+function SelectInput(props: { kind: MediaDeviceKind }) {
+  // eslint-disable-next-line solid/reactivity
+  const kind = props.kind;
```
Dadadah: This violates the code style guidelines. Please don't destruct props.

packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
```diff
@@ -18,101 +14,73 @@ export function VoiceInputOptions() {
...
-    useMediaDeviceSelect({
-      kind: "audioinput",
-    });
+    useMediaDeviceSelect({ kind });
 
-  const activeId = () =>
-    (activeDeviceId() === "default"
-      ? state.voice.preferredAudioInputDevice
-      : undefined) ?? activeDeviceId();
+  const setKey =
```
Dadadah: I may be mistaken here, but I think these consts (setKey, icon and title) are not reactive. I think they should be accessors?

packages/client/components/app/interface/settings/user/Language.tsx
Dadadah, line 1: Your edits to this file have made the language select lose reactivity if the user changes their preferred locale on the browser side. The code before was reactive by using memos and accessors, Shouldn't it have remained that way? Specifically `initLang`, `prefLang`, `langOpts`, and `langIds`.

Pecacheu: Initially it was to avoid an infinite reactivity loop that was happening, basically you change the value of the Select in onUpdate and then the input value changes too, which causes onUpdate to fire again, and so on-

However I believe I've fixed that loop from being possible inside the code of the Select, but might need to do some more testing on that to make sure the loop is impossible.

packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
Pecacheu: This is due to a combo of _both_ below issues so both need to be addressed first

packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
Pecacheu: Good point, this is unrelated to the whole reactivity loop issue. We're not using reactivity here since the prop value is hard-coded in VoiceInputOptions, but it wouldn't hurt just in case. I can make them a function or a memo instead to address that.

Pecacheu: Okay nvm I have no idea how to do this without making an absolute mess of the code lol
Basically if you change props.kind the ENTIRE thing needs to be re-rendered from scratch

Pecacheu:
Ok it's officially fully reactive (and tested) now, aside from the Languages menu which appears to not support reactivity due to the underlying `window.navigator` call being non-reactive. Thanks for the help @Dadadah

packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
```diff
@@ -18,101 +14,74 @@ export function VoiceInputOptions() {
...
  */
-function SelectMicrophone() {
-  const { t } = useLingui();
+function SelectInput(props: { kind: MediaDeviceKind }) {
   const state = useState();
-  const { activeDeviceId, devices, setActiveMediaDevice } =
-    useMediaDeviceSelect({
-      kind: "audioinput",
-    });
+  const media = createMemo(() => useMediaDeviceSelect({ kind: props.kind }));
```
gabrielfordevelopment, line 37: **Suggested code:**
```tsx
function SelectInput(props: { kind: MediaDeviceKind }) {
  const state = useState();
  const media = useMediaDeviceSelect({ kind: props.kind });
}
```


`createMemo` is for pure computations. Calling a hook with side effects inside it can silently leak signals, effects, and event subscriptions if the memo ever re-runs. Calling the hook directly at the component top level is the standard SolidJS pattern and eliminates this risk.

packages/client/components/ui/components/design/CategoryButton.tsx
```diff
@@ -351,9 +344,117 @@ export function CategoryCollapse(props: CollapseProps) {
       </Switch>
     </Details>
   );
-}
+};
+
+export type CatSelOption = Omit<
```
gabrielfordevelopment: **Suggested code:**

```tsx
export type CategorySelectOption = Omit<...> & (...);
```

Other exported types in the `design/` layer use full names (`DialogProps`, `DialogAction`, `DateTimePickerMode`). `CatSelOption` is abbreviated inconsistently. `CategorySelectOption` is more discoverable and follows existing conventions.

packages/client/components/ui/components/design/CategoryButton.tsx
```diff
@@ -351,9 +344,117 @@ export function CategoryCollapse(props: CollapseProps) {
...
+        shortDesc: JSX.Element;
+      }
+  );
+
+type SelectProps<T extends string> = Omit<
+  ComponentProps<typeof CategoryButton>,
+  "onClick" | "children" | "description"
+> & {
+  title?: JSX.Element;
+  options: { [k in T]: CatSelOption };
```
gabrielfordevelopment: https://github.com/stoatchat/for-web/pull/1085/changes#r3066281735

packages/client/components/ui/components/design/index.ts
```diff
@@ -31,7 +31,7 @@
 export { Avatar } from "./Avatar";
 export { Badge } from "./Badge";
 export { Button } from "./Button";
-export { CategoryButton } from "./CategoryButton";
+export { type CatSelOption, CategoryButton } from "./CategoryButton";
```
gabrielfordevelopment: https://github.com/stoatchat/for-web/pull/1085/changes#r3066281735

packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
Dadadah: When it was like this, it was not reactive to kind. We had a discussion about this, and I think the hook is broken. Do you know how to fix it? You should try in your environment.

gabrielfordevelopment: Fixing the hook is **out of scope for this PR** — should be a separate PR against `packages/solid-livekit-components`. The `createMemo` workaround is acceptable for now.

packages/client/components/ui/components/design/CategoryButton.tsx
Dadadah: I agree that it should be renamed

Pecacheu: Done!

PR #1086, opened 2026-04-02
feat: Video device option in voice settings

Pecacheu:
Built on #1085 (now merged)

### DIFF
--- a/packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
+++ b/packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
@@ -22,7 +22,7 @@ export function VoiceInputOptions() {
       <CategoryButton.Group>
         <SelectInput kind="audioinput" />
         <SelectInput kind="audiooutput" />
-        {/* <SelectInput kind="videoinput" /> TODO O.o */}
+        <SelectInput kind="videoinput" />
       </CategoryButton.Group>
       <VolumeSliders />
     </Column>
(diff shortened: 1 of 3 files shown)


packages/client/components/app/interface/settings/user/voice/VoiceInputOptions.tsx
Dadadah: I guess the video device list doesn't return with a default, but if it's unset it'll use the system's default device. This results in the video device being unselected initially. You can change this line to insert a default regardless of whether one is found and it will make Default show up. Picture to explain what I mean:

Image

```suggestion
    opts.default = { title: d?.label ?? "Default" };
```

After change:

Image

Pecacheu: Oh huh, didn't know that was how to make it use the default. I can add that

Pecacheu: Hmm... Gave it a try, but no dice unfortunately. Getting this error if I try to set the video device to either `""` or `"default"` (this occurs once turning on the camera after setting it)


So while adding a "Default" item to the options list is easy, making it do anything is not. But maybe there's something in livekit I'm missing.

Dadadah: actually, how does this PR even pass the preferred camera to livekit?

Pecacheu: Hmm actually you're right that there's something missing.

For the most part it works the same way the audio input/output are passed to livekit. When the selection changes, onUpdate is called and the selected key is double-checked just to make sure the device with that ID actually exists (a check that I was overriding in order to perform the above test), the user's new selection is saved to the corresponding voice state store, and finally, `setActiveMediaDevice` is called to tell livekit to use it.

Buuuuut this only works until the app is reloaded. What's missing is that those voice state stores are also read when calling `new Room` in state.tsx, so that livekit sets the same audio devices again even after a reload. So I need to add in setting the video device to `this.#settings.preferredVideoDevice` in there somewhere.

As for the default device issue, I believe that the reason trying to set it to "default" is failing is that livekit might not support that. When it sends over the list of audio inputs/outputs, there is always device with the ID "default" in the list. But when it sends over the video devices, there is no such device :(

Dadadah: I'm wondering if you can just set the value to an empty string or something. It just needs to show "Default"

Pecacheu: Already tried empty string. I can try null or undefined and see if that works? (Basically would just require an extra if statement in onUpdate that if type is video and name is default, set to undefined instead)

Pecacheu: Think I managed to get it working with undefined. Tho it's not possible to call setActiveMediaDevice with undefined, so basically if you set it to default it will continue using the previous video device until you reload the window, when livekit will, in the absence of a specified video device, use the default again

Dadadah (approved):

PR #1087, opened 2026-04-02
feat: Remember mic mute setting

Pecacheu:
I think this one is self-explanatory lol

- `main` 
  - \#1087 :point\_left:

### DIFF
--- a/packages/client/components/rtc/state.tsx
+++ b/packages/client/components/rtc/state.tsx
@@ -48,10 +48,7 @@ class Voice {
   #setState: Setter<State>;
 
   deafen: Accessor<boolean>;
-  #setDeafen: Setter<boolean>;
-
   microphone: Accessor<boolean>;
-  #setMicrophone: Setter<boolean>;
 
   video: Accessor<boolean>;
   #setVideo: Setter<boolean>;
@@ -87,13 +84,8 @@ class Voice {
     this.state = state;
     this.#setState = setState;
 
-    const [deafen, setDeafen] = createSignal<boolean>(false);
-    this.deafen = deafen;
-    this.#setDeafen = setDeafen;
-
-    const [microphone, setMicrophone] = createSignal(false);
-    this.microphone = microphone;
-    this.#setMicrophone = setMicrophone;
+    this.deafen = () => voiceSettings.deafen;
+    this.microphone = () => voiceSettings.micOn;
 
     const [video, setVideo] = createSignal(false);
     this.video = video;
@@ -145,9 +137,6 @@ class Voice {
       this.#setRoom(room);
       this.#setChannel(channel);
       this.#setState("CONNECTING");
-
-      this.#setMicrophone(false);
-      this.#setDeafen(false);
       this.#setVideo(false);
       this.#setScreenshare(false);
     });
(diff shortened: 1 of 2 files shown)


gabrielfordevelopment:
Could we persist all the voice call settings?
- Microphone mute bool
- Headphone mute bool
- Volume % of individual users
- (Upcoming) stream audio mute bool
- (Upcoming) stream volume %

packages/client/components/rtc/state.tsx
```diff
@@ -161,7 +163,9 @@ class Voice {
       !room.localParticipant.isMicrophoneEnabled,
     );
 
-    this.#setMicrophone(room.localParticipant.isMicrophoneEnabled);
+    const on = room.localParticipant.isMicrophoneEnabled;
+    this.#settings.micOn = on;
+    this.#setMicrophone(on);
```
Dadadah: Since we're storing this in the state now, I think we should get rid of the microphone state tracking in the rtc state. I don't think it makes sense to have two sources of truth for this, and it introduces complexity in keeping the two states equal. Rather than keeping both states, what if instead the `microphone` accessor on the rtc state just relayed the voice settings state accessor to keep it backwards compatible? I believe the only reason the state is tracked here is to be reactive, otherwise we could just use room.localParticipant.isMicrophoneEnabled.

packages/client/components/rtc/state.tsx
```diff
@@ -161,7 +163,9 @@ class Voice {
       !room.localParticipant.isMicrophoneEnabled,
     );
 
-    this.#setMicrophone(room.localParticipant.isMicrophoneEnabled);
+    const on = room.localParticipant.isMicrophoneEnabled;
```
Dadadah: It wasn't clear to me on first glance that `on` would be set to the new value, maybe instead of setting it here, it can be set above on line 162 to `!room.localParticipant.isMicrophoneEnabled`. That new variable can then be used in both the setter for the room and the setter for the settings state.

Dadadah (changes requested):

Pecacheu:
@gabrielfordevelopment "Volume % of individual users" is actually already in main. I can add deafen tho.

packages/client/components/rtc/state.tsx
Pecacheu: This is a good point lol

packages/client/components/rtc/state.tsx
Pecacheu: Actually that was my first instinct too but from testing I quickly realized why it was done like that, it is because if the mic fails to actually enable, then when the variable is read back after being set to true, livekit will have gone "nuh uh" and set it back to false again, thus the new saved status reflects the fact that mic failed to turn on.

However that `on` var can be deleted anyhow given the other change of making the mic and deafen accessors forward to the Voice store.

packages/client/components/rtc/state.tsx
```diff
@@ -103,26 +95,25 @@ class Voice {
...
-          if (this.#settings.noiseSupression === "enhanced") {
-            track?.audioTrack?.setProcessor(
-              new DenoiseTrackProcessor({
-                workletCDNURL: CONFIGURATION.RNNOISE_WORKLET_CDN_URL,
-              }),
-            );
-          }
-        });
+        room.localParticipant
+          .setMicrophoneEnabled(this.#settings.micOn)
```
Dadadah, line 148: I didn't think of bringing this up earlier - should we catch this promise on fail to set micOn to false?

Dadadah: Nevermind this is a much more involved than I initially thought.

PR #1090, opened 2026-04-04
feat: new voice ui better suited for video and screen share with focus feature

### LINKED ISSUE #817: Weird UI bug on vc
### What happened?

On the new version of Stoat i got a pretty weird ui glitch, when i join a vc i see a scroll bar on the right that's not supposed to be here

### LINKED ISSUE #1067: refactor: Rebuild the Call UI
Rebuild the call UI to better host screen sharing and video functionality.

Dadadah:
New voice UI better suited for video streaming and screen sharing with a focus feature. Collaboration between @Pecacheu and I.

Supersedes #1061 (Thanks @misofist)

Closes #1067
Fixes #817

This PR implements a new Voice UI that is better suited for video streaming. It increases the size of the tiles so that you can actually see multiple camera streams. It also adds the ability to full screen the voice UI as a whole instead of just a single video stream.

This PR imports @solid-primatives/resize-observer. I am very much not a fan of AutoResizer as it adds an additional div. Resize observer does the same thing but without the extra div.

Demo:

[Stoat\_Demo\_Full.webm](https://github.com/user-attachments/assets/44c7c469-d2b0-4331-9ba4-21f7b109abe7)

There was also a small change to the voice call pip, Demo:

[Stoat\_Demo\_Pip.webm](https://github.com/user-attachments/assets/86b726a6-d725-4b91-88f6-9a514109dbd1)

Summary of changes:

- Voice tiles now fill the voice call container
- Voice tiles are now reactive
- Voice container no longer has scrollbar unless there are many users connected (#817)
- Voice container now has properly styled scrollbars
- Video and screenshare now always fit inside the tiles when not focused
- Add ability to choose one tile as the focus
- Video and screenshare take up significantly more room when focused
- Add ability to hide non-focused participants
- Add ability to fullscreen the voice call container
- Add fullscreen button on bottom right
- Screenshare tiles are adaptive to streamed resolution when focused
- Move voice status to bottom left
- Make `Connected` fade out in the voice status. All other statuses do not fade out
- Fix cursor always being a grab cursor on the voice box
- Refactor active call component to only have a single tile instead of user and screen share tiles being separate
- Import @solid-primatives/resize-observer

Known issues:

- Voice status text is clipped by the controls on small screens
  - May require a sizable refactor to fix, I'd like to do this in another PR after this one is merged.
- Flickers when resizing
  - Existed before. Likely requires refactoring the voice call card holder. I'd like to refactor the holder in another PR after this one is merged.

* `main` 
  - \#1090 :point\_left:


packages/client/components/ui/components/features/voice/callCard/VoiceCallCardStatus.tsx
```diff
@@ -9,32 +7,76 @@ import { Symbol } from "@revolt/ui/components/utils/Symbol";
...
-        </Match>
-        <Match when={voice.state() === "DISCONNECTED"}>
-          <Symbol>wifi_tethering_error</Symbol> <Trans>Disconnected</Trans>
-        </Match>
-        <Match when={voice.state() === "RECONNECTING"}>
-          <Symbol>wifi_tethering</Symbol> <Trans>Reconnecting</Trans>
-        </Match>
-      </Switch>
+      <Symbol>{symbol()}</Symbol>{" "}
+      <FadeOut fade={voice.state() === "CONNECTED"}>{text()}</FadeOut>
```
amycatgirl, line 43: I think that on smaller viewports, this text should be hidden or moved into a popover. See attachment

Image

amycatgirl:
oh, also when a video stream is focused, it should take more of the available space inside of the card container because text being streamed can become unreadable. I don't quite remember how it looks like on other platforms such as Slack or Discord, but behaviour should be at least similar to what they do

packages/client/components/ui/components/features/voice/callCard/VoiceCallCardStatus.tsx
Pecacheu: Yeah this was the only real outstanding issue on this one but we agreed it's relatively minor for now since mobile UI in general is lacking in the app and this is 100% usable, just not pretty. (Plug for #835 tho :p)

Additionally, Dadadah made it so that the fullscreen button takes priority over the status text, i.e. the bar moves to the left instead of right, just to go down to the smallest size possible on this current design.

Pecacheu:
> The PIP UI should also be improved

@amycatgirl You're in luck, I'm working on that, but it shall come later in it's own PR :3

packages/client/components/ui/components/features/voice/callCard/VoiceCallCardStatus.tsx
Dadadah: It's a pretty sizable refactor to get this to work nicer. I figured that this would be acceptable for now, and maybe we could refactor it in another pr. I can do it in this PR if it doesn't pass UI review.

Dadadah:
> oh, also when a video stream is focused, it should take more of the available space inside of the card container because text being streamed can become unreadable. I don't quite remember how it looks like on other platforms such as Slack or Discord, but behaviour should be at least similar to what they do

Can you go into more detail about this? The stream should be almost the full screen when fullscreened. I don't think it's realistic to read the font text without being in fullscreen.

gabrielfordevelopment:
Everything have a nice animation to open/close.
Except for the Screen Focus change, it just happens instantly without any transition.
It works and is usable, but feels a bit out of space to give a bit of animation to everything, but the screen change is instant.
It might be a big refactor, or hard to change, but wanted to leave this comment here.
Other than that LGTM!

I believe that as we use the screen share feature and get more experience we will see what needs to be adjusted or improved. But for now, this looks great


Oh and one small detail, this should be either "Confirm" or "Select" and not "Share". You only select it here, and click "Share" **after** this selection is done:
image

Then you click "Share" here, that is good, but the first button ↑ should be either "Confirm" or "Select" above
image

Dadadah:
> Oh and one small detail, this should be either "Confirm" or "Select" and not "Share". You only select it here, and click "Share" **after** this selection is done: image
> 
> Then you click "Share" here, that is good, but the first button ↑ should be either "Confirm" or "Select" above image

These UI's are provided by the browser and my OS (Linux with Wayland). They're not built by Stoat.

amycatgirl:
> Can you go into more detail about this? The stream should be almost the full screen when fullscreened. I don't think it's realistic to read the font text without being in fullscreen.

In fullscreen it is readable, yes. I was indeed referring to the non-fullscreen ui.

In fullscreen, it looks perfect, since the card fills the entire screen. 

I was thinking something similar to this mockup i made?

image
(Excuse my phone's broken lens)

The screen-share/video feed fills the entire card and keeps the buttons visible, while the participant's name and muted status is moved to the top right, and the functionality to show other participants stays the same.

Pecacheu:
> The screen-share/video feed fills the entire card and keeps the buttons visible, while the participant's name and muted status is moved to the top right, and the functionality to show other participants stays the same.

I think this is doable, eg. if fullscreen, in focused view, AND with the other participants bar hidden, the bottom controls could fade out when there's no mouse movement like the controls in a full screen video player.

Dadadah:
> > The screen-share/video feed fills the entire card and keeps the buttons visible, while the participant's name and muted status is moved to the top right, and the functionality to show other participants stays the same.
> 
> I think this is doable, eg. if fullscreen, in focused view, AND with the other participants bar hidden, the bottom controls could fade out when there's no mouse movement like the controls in a full screen video player.

I think we should avoid doing this here. Blocking the video that much is going to be undesirable in some circumstances, and hiding the controls is not an option right now due to lack of mobile device detection. Perhaps this can be investigated again once we have a standard for detecting device features?

amycatgirl:
makes sense

though, i could see some users wanting to see the stream more clearly without having to go into fullscreen

i guess y'all can leave that for later, or unless either another user complains about this or UI/UX has something planned for this

packages/client/components/ui/components/features/voice/callCard/VoiceCallCardStatus.tsx
lemte: image

I'm just an enthusiast, but I saw the discussion and wanted to suggest an idea. ^^
I explored a cleaner version of the call interface focusing on smaller screens. I removed the text and used only icons to reduce visual clutter and improve responsiveness. The actions are grouped (audio on the left, video/screen on the right), with the end button centered for better positioning.

Dadadah: Thanks for the input, that does look nice but a voice control refactor is out of scope for this PR.

PR #1091, opened 2026-04-04
feat: "Enable devtools" checkbox for Desktop

Taureon:
Adds a checkbox that gives you a warning about the dangers of enabling devtools, but ultimately lets you enable devtools for a day.

Current issues before it can be opened:
~~1. If the checkbox itself is clicked, the modal appears twice for some reason~~
2. The checkbox itself still toggles even if the modal was just activated
3. No warning text lol, there's just a placeholder
4. This bricks the settings menu for some reason after you go into a submenu

Relies on https://github.com/stoatchat/for-desktop/pull/201

Also I'm DCO-ing this after it's finished


Taureon:
things to warn about evil people doing:
- pasting viruses into the console
- screenshotting the network tab
- screensharing the network tab
todo: figure out if i should add more ig

PR #1092, opened 2026-04-04
fix: Add button action timeout to solve click-through issue on touch devices

### LINKED ISSUE #798: In Chrome, touch events on some buttons propagate through after dialogs close

Pecacheu:
Fixes #798

At long last, I found a simple fix that works on Chrome mobile without breaking it on Firefox mobile!

This also fixes another button bug on mobile where when the virtual keyboard is open, pressing a button does not fire as it closes. This is likely a Solid Aria bug, but adding `onTouchEnd` fixes it for now, and the debounce prevents multiple event firings.

### DIFF
--- a/packages/client/components/ui/components/design/Button.tsx
+++ b/packages/client/components/ui/components/design/Button.tsx
@@ -122,6 +122,7 @@ export function Button(props: Props) {
 
   const [btn, noBtnRest] = splitProps(rest, ["onPress"]);
 
+  //Eslint being silly, this is reactive
   // eslint-disable-next-line @typescript-eslint/no-explicit-any, solid/reactivity
   const onPress = debounce((e: any) => btn.onPress?.(e), 100),
     btnRest = mergeProps(noBtnRest, { onPress });
--- a/packages/client/components/ui/components/design/Button.tsx
+++ b/packages/client/components/ui/components/design/Button.tsx
@@ -5,6 +5,7 @@ import { AriaButtonProps, createButton } from "@solid-aria/button";
 import { cva } from "styled-system/css/cva";
 
 import { debounce } from "@revolt/common";
+
 import { Ripple } from "./Ripple";
 import { typography } from "./Text";
 
@@ -90,7 +91,7 @@ export function Button(props: Props) {
     "role",
   ]);
 
-  const [style, noSty] = splitProps(propsRest, [
+  const [style, rest] = splitProps(propsRest, [
     "bg",
     "size",
     "shape",
@@ -119,17 +120,18 @@ export function Button(props: Props) {
     ),
   );
 
-  const [btn, noBtnRest] = splitProps(noSty, ["onPress"]);
+  const [btn, noBtnRest] = splitProps(rest, ["onPress"]);
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any, solid/reactivity
   const onPress = debounce((e: any) => btn.onPress?.(e), 100),
-    rest = mergeProps(noBtnRest, { onPress });
+    btnRest = mergeProps(noBtnRest, { onPress });
 
-  const { buttonProps } = createButton(rest, () => ref);
+  const { buttonProps } = createButton(btnRest, () => ref);
   return (
     <button
       {...passthrough}
       {...buttonProps}
+      onTouchEnd={onPress}
       ref={ref}
       class={button({
         shape: shape(),
--- a/packages/client/components/ui/components/design/Button.tsx
+++ b/packages/client/components/ui/components/design/Button.tsx
@@ -1,9 +1,10 @@
-import { Show, createRenderEffect, on, splitProps } from "solid-js";
+import { Show, createRenderEffect, mergeProps, on, splitProps } from "solid-js";
 import { JSX } from "solid-js/jsx-runtime";
 
 import { AriaButtonProps, createButton } from "@solid-aria/button";
 import { cva } from "styled-system/css/cva";
 
+import { debounce } from "@revolt/common";
 import { Ripple } from "./Ripple";
 import { typography } from "./Text";
 
@@ -89,7 +90,7 @@ export function Button(props: Props) {
     "role",
   ]);
 
-  const [style, rest] = splitProps(propsRest, [
+  const [style, noSty] = splitProps(propsRest, [
     "bg",
     "size",
     "shape",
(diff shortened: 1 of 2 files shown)


packages/client/components/ui/components/design/Button.tsx
```diff
@@ -1,9 +1,10 @@
-import { Show, createRenderEffect, on, splitProps } from "solid-js";
+import { Show, createRenderEffect, mergeProps, on, splitProps } from "solid-js";
 import { JSX } from "solid-js/jsx-runtime";
 
 import { AriaButtonProps, createButton } from "@solid-aria/button";
 import { cva } from "styled-system/css/cva";
 
+import { debounce } from "@revolt/common";
```
Dadadah: Can you reorder these imports to have  the `@revolt` import separate from the local imports to match the order found in other files?

packages/client/components/ui/components/design/Button.tsx
```diff
@@ -118,6 +119,12 @@ export function Button(props: Props) {
     ),
   );
 
+  const [btn, noBtnRest] = splitProps(noSty, ["onPress"]);
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, solid/reactivity
```
Dadadah: Could you explain why these eslint directives are here? The no-explicit-any isn't that big of a deal, but solid/reactivity is a bit bigger deal, what loses reactivity here?

Dadadah (changes requested):

packages/client/components/ui/components/design/Button.tsx
Pecacheu: Edit: Okay I originally said "Just copying the debounce code from IconButton" but it turns out that is from another PR that isn't merged yet.

In which case, I'm going to change it in both places to ensure consistency.

Pecacheu: Okay I'm struggling to see how it's not reactive actually, the place where it throws the warning is in the `=>`. But if onPress changes, then won't the debounce func just call the new one?

packages/client/components/ui/components/design/Button.tsx
Pecacheu: Fixed

packages/client/components/ui/components/design/Button.tsx
Dadadah: `debounce` is given the initial value of `onPress`, and I don't think it will change if the props change. I think you may have to create an effect that updates the `rest` object with a new debounced `onPress`.

Pecacheu: Alright, I put together a test that updates a signal set as the onPress listener with randomly generated functions, so I can now confirm that the warning **liesssss** and it is 100% reactive. Which makes sense given the use of splitProps and mergeProps along with debounce being set to an intermedite function that always grabs the latest onPress.

Dadadah (approved):

packages/client/components/ui/components/design/Button.tsx
chrishultin: Can you go ahead and add a note/comment explaining that the warning is invalid/inaccurate?

Pecacheu: > Can you go ahead and add a note/comment explaining that the warning is invalid/inaccurate?

Good idea, will do

Dadadah (changes requested):
I've taken another look at this, and I think the onTouchEnd change should be removed. It changes the behavior of the buttons on mobile, and I don't think it's actually an aria bug. The issue is due to the modal moving when the keyboard is closed. You can view a similar situation at https://material-web.dev/components/button/#usage and notice that the buttons don't move, therefore it works.

Pecacheu:
@Dadadah Actually, not quite- From my extensive testing on different devices, including Chromium-based and Firefox on several Android phones and tablets, as well as Safari on (simulated) iOS, buttons- including those that move when the touch keyboard closes- trigger just fine and do not have click-though issues in basically anything *except* for Stoat web _(Note: Firefox does not experience the click-through bug to begin with, but does intermittently experience the keyboard bug.)_ This problem also doesn't occur on mobile in any of the webapps I've made using Bootstrap modals with keyboard-interactable fields, which is an extremely similar UI situation to what causes the issue in Stoat web.

That's what has me convinced it's a bug in aria button specifically. Pretty much the only explanation for this difference is the _hubris_ of the library to totally override the browser's native HTML5 button event handling (tbh this is why when I make webapps I opt for as much native elements/behavior as possible.)

So far, `onTouchEnd` is the only way I've found to get the aria buttons to behave. If you find something else that works better, I'd take it, but I see no issue here. Fixes both bugs on Chromium/Safari, the one intermittent bug on Firefox, and doesn't break anything on either in the process. On that note, when you say "it changes the behavior of the buttons on mobile" what do you mean (and which device + browser are you seeing the issue on?) In all my tests, the buttons now behave identically on desktop and mobile. As mentioned, the debounce code prevents duplicate event listeners from firing... plus, it does double-duty preventing an accidental fat-finger double-press in rapid succession, which is a lot easier to do on a touchscreen than with a mouse, trust me.

**Note:** A nice spot in the app to test the change is under `Settings -> Sessions -> Rename`. A modal with two buttons and a textbox to open the keyboard, so you can test both bugs here.

Dadadah:
I think my main concern here is that we're implementing a "bugfix" for something that we have no documentation of other users of our stack reporting. It's possible it's a solid-aria bug, and if so, we should open an  issue upstream. It's also possible, though, that it's an implementation issue. If it's an implementation issue I don't want to implement a bandaid on top of it. I think it's worth it to investigate this issue further. I'm going to spend some time investigating it and see if I can get it to replicate with just a basic aria button.

Pecacheu:
Ok, just make sure you test using Chrome/Chromium since it seems to reproduce the issue more reliably than Firefox (which only sometimes does it?)

Dadadah:
Opened an issue upstream:
https://github.com/solidjs-community/solid-aria/issues/84

packages/client/components/ui/components/design/Button.tsx
```diff
@@ -118,7 +118,14 @@ export function Button(props: Props) {
     ),
   );
 
-  const { buttonProps } = createButton(rest, () => ref);
+  const [btn, noBtnRest] = splitProps(rest, ["onPress"]);
+
+  //Emulate delay of native onClick (solves click-through issue)
```
Dadadah: Please reference the issue I linked on the solid-aria repo in the comments here so it can be referenced in the future.

Pecacheu: Fixed, also changed to 32ms due to the Safari thing

Dadadah (approved):

PR #1102, opened 2026-04-07
refactor: Voice Call PIP Card

Pecacheu:
- Added missing tooltips for all the action buttons
- Massive performance improvement for dragging & dropping the PIP window around
- Compatibility with touchscreen dragging as well as mobile layout in #835
- Mini video preview when focused view from #1090 is active
- Dim background behind muted icons to ensure they are visible on both dark and light avatars

Without focused video: 

With focused video: 

Dragging (no change to animations)

<https://github.com/user-attachments/assets/7ccd5328-7451-41d0-9d34-6376107b615f>

Slightly tweaked drop animation, at suggestion of @Dadadah, before vs after

<https://github.com/user-attachments/assets/e8187315-940f-4bbf-b53f-487cc92bcf5b>

- `main` 
  - \#1102 :point\_left:

### DIFF
--- a/packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
+++ b/packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
@@ -1,13 +1,9 @@
 import {
   JSX,
-  Match,
   Show,
-  Switch,
-  batch,
   createContext,
   createEffect,
   createSignal,
-  on,
   onCleanup,
   onMount,
   useContext,
@@ -16,7 +12,6 @@ import { Portal } from "solid-js/web";
 
 import { AutoSizer } from "@dschz/solid-auto-sizer";
 import { Channel } from "stoat.js";
-import { css } from "styled-system/css";
 import { styled } from "styled-system/jsx";
 
 import { InRoom, useVoice } from "@revolt/rtc";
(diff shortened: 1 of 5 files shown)


packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
```diff
@@ -3,284 +3,240 @@ import {
...
+  const [mode, setMode] = createSignal<Mode>();
+  const [info, setInfo] = createSignal<Info>();
+  let ref: HTMLDivElement | undefined;
+
+  let events: AbortController | null,
+    tid = 0,
+    ofsX = 0,
+    ofsY = 0;
+
+  function touchToMouse(e: MouseEvent | TouchEvent, down = false) {
```
Dadadah: Could this, instead of returning a boolean, return a partial TouchEvent with just the info needed for processing? That would get rid of the need to use ts-expect-error here. Calling it could be like `const partialEvent = touchToMouse(e)`. It would also get rid of the `e as MouseEvent` in the mouseDown function below.

Dadadah: Should this function be moved into a utility library somewhere to allow this across the app?

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
```diff
@@ -3,284 +3,240 @@ import {
...
+  function mouseUp(e: MouseEvent | TouchEvent) {
+    if (!touchToMouse(e)) return;
+    const sty = ref!.style,
+      pos = ref!.getBoundingClientRect(),
+      left = (e as MouseEvent).clientX - ofsX + pos.width / 2 < outerWidth / 2,
+      top = (e as MouseEvent).clientY - ofsY + pos.height / 2 < outerHeight / 2;
+
+    sty.transition = "all .2s cubic-bezier(0, 1.67, 0.85, 0.8)";
+    setFloat(left ? (top ? "tl" : "bl") : top ? "tr" : "br");
+    setTimeout(() => (sty.transition = ""), 1);
```
Dadadah, line 82: I haven't tested this, but I think this would result in the transition being lost if you click, drop, then click within 1 second, wait 1 second and drag to a new spot.

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
```diff
@@ -3,284 +3,240 @@ import {
...
+      floating: { cursor: "grab" },
+      moving: {
+        cursor: "grabbing",
+        transition: "none",
+      },
+    },
+  },
+  compoundVariants: [
+    {
+      mode: ["floating", "moving"],
```
Dadadah, line 170: I don't think this can be reached? Mode is a union type of "floating" and "moving", therefore it can't be both.

Dadadah (changes requested):
Please remove the TODOs and dead code regarding your other PRs. If this were to get merged but the other PRs weren't these TODOs would require a PR to be removed. These lines are unrelated to this PR.

Could you show more pictures of the pip, including without focus, with just voice users (no video), and a screencast of the pip being moved around the screen?

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Pecacheu: Good idea, just curious what you think would be a good in the file-tree for it. I could see some touch-related methods used in #835 going there too, actually.

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Pecacheu: Actually the intent was an OR between floating and moving rather than AND, aka it happens for either one. Which I _think_ is working correctly but I can double-check later

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Pecacheu: It's actually only 1ms rather than 1s, so it'd be VERY difficult to pull that off in the real world, but yeah you are correct that sequence of events timed frame-perfect would break it.

If there's a better way to do it I'm open to it, the purpose of the timeout is basically just to remove the transition property immediately on the next render pass- This works because the animation will complete anyway. I'm not sure if using a mechanism like requestAnimationFrame might be better for this than setTimeout.

Dadadah: oh yeah I totally read that wrong. I'm not sure the best way to do it honestly, and if it works then it's probably fine. Could you leave a comment explaining why the timeout is there? What you said here is probably fine `//Remove the transition property immediately on the next render pass`

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Dadadah: Actually looking at this again, can you just use the [PointerMove event](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointermove_event) instead? This would require applying `touch-action: none` to the pip, which is probably something we want anyway right?

Pecacheu: Yeah good idea, that one didn't even occur to me because last time I checked out the newer pointer events was several years back, when support was not widespread (eg. on iOS and Firefox Mobile) and the implementation was not consistent, but since 2020 it now is (or so the API support sites say at least, but we should def test on different devices to be sure)

Damn can't believe it has been 6 years since 2020 lol

Pecacheu: Okay I was able to switch to pointer events- It did indeed simplify the code

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
```diff
@@ -3,282 +3,221 @@ import {
...
-      setState({
-        type: "floating",
-        corner: "bottom-right",
-      });
-    }
+  function mouseUp(e: PointerEvent) {
+    if (e.pointerId !== pid) return;
+    const sty = ref!.style,
+      pos = ref!.getBoundingClientRect(),
+      left = e.clientX - ofsX + pos.width / 2 < outerWidth / 2,
```
Dadadah: I think these should use innerWidth. Outerwidth makes it not work as expected with dev tools is open, or in responsive mode on desktop. Inner width and and inner height seem to work as expected.

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
```diff
@@ -3,282 +3,221 @@ import {
...
+      top = e.clientY - ofsY + pos.height / 2 < outerHeight / 2;
+
+    sty.transition = "all .2s cubic-bezier(0, 1.67, 0.85, 0.8)";
```
Dadadah: I think the cubic transition is causing the voice pip to partially exit the window on larger screen sizes. Might be best to not use cubic here as any cubic function will result in this happening. This applies to all animations regarding the pip. If you can figure out how to get it to animate cubic without doubling back that also works.

packages/client/components/ui/components/features/voice/callCard/VoiceCallCardPiP.tsx
```diff
@@ -48,19 +68,58 @@ function ConnectedUser() {
 
   return (
     <UserIcon speaking={isSpeaking()}>
-      <Avatar size={24} src={user().avatar} fallback={user().username} />
+      <Avatar
+        size={24}
+        src={user().avatar}
+        fallback={user().username}
+        shape="square"
```
Dadadah, line 75: Can you explain why you added square avatars, and why removing this line of code doesn't seem to do anything?

Dadadah (changes requested):
Thanks for addressing the first review. I think after these three issues are addressed we can get this lined up for a UI/UX review.

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Pecacheu: Haha tbh making sexy animated transitions was never my specialty, there's a reason I went into drawing still art but not animation.

These cubic bezier animations are already the ones in use in the current release build and I didn't alter their behavior (just moved where they are in the code while rewriting the drag), but I'd be happy to update them here if someone comes up with nicer looking ones (I believe there are some online tools that can help with editing the curves with live preview). I think they look alright tho.

packages/client/components/ui/components/features/voice/callCard/VoiceCallCardPiP.tsx
Pecacheu: It's a weird subtly because I was being pedantic. By moving the circle mask out of the avatar and into the element outside of it, I can make the masking for the mic-mute overlay pixel-perfect. Previously, it was a pixel or two off because we were placing an SVG masked round over another SVG separately masked round. Now it's two unmasked SVGs inside an element that masks both round simultaneously.

An alternate solution would be adding a 'muted' option to Avatar and placing the mute symbol inside the code for it, but that felt kind of random to insert into the Avatar code.

The reason you're not seeing a change between the two is because we're talking about a pixel-level imperfection you'd only see if you're using a high-DPI display like my 4K monitor lol

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Pecacheu: Tbh I was using outerWidth here because the previous code for the mouse-up listener used outerWidth. But yeah, innerWidth is probably better, I'll update and test when I have time

packages/client/components/ui/components/features/voice/callCard/VoiceCallCard.tsx
Dadadah: Change the second number to 1.5 and it will keep the little wiggle but not wiggle out of view. You can test the params using this website: https://cubic-bezier.com/#0,1.5,.85,.8

Pecacheu: Added in! Added a video of the difference too so UI/UX can see if they like it or not

Dadadah (approved):

infi (approved):

Issue #1070, opened 2026-03-28
feat: Add sound to screen sharing


Dadadah:
Add the ability to stream sound when screen sharing.

Most platforms support this. Requires the ability to mute and/or change the volume of screen shares.

ThomasJRyan:
# Description

Add the ability to stream the computer's sound when screen sharing. In the event of screen sharing, share the sound of everything coming from the computer. In the event of window sharing, share only the sound of the window.

There is currently discussion about looking into https://github.com/Vencord/venmic to see how this project handles sound in screen sharing. If this is used, ensure it abides by the MPL-2.0 license.

# DOD

- Stream the user's desktop environment's sound
  - If sharing screen, stream all sound
  - If sharing window, stream only that window's sound

# Considerations

- Ensure all open source licenses of reference projects are followed

Issue #1068, opened 2026-03-28
feat: Windows screen picker for Desktop


Dadadah:
Add a screen picker for the electron app on Windows.

On Linux and MacOS, the OS's handle screen picking functionality for electron apps. On Windows, there should be a screen picker available to choose which window to stream. This can possibly be built in/built with a stream settings UI.

ThomasJRyan:
# DOD

- A screen/window picker available to the application
  - Should be able to detect screens and windows running on the computer
- Leverage each OS' functionality to pick screens/windows
- A stream settings UI (see #1069)

# Considerations

- All UI/UX must be approved by @stoatchat/ui-ux

PR #1111, opened 2026-04-11
feat: Screen share settings UI modal

### LINKED ISSUE #1069: feat: Stream Settings UI
Add a stream settings UI.

There are many configuration options available for screen and video sharing. Right now, they are all hardcoded to default values. This UI should be able to configure the following:

- Stream resolution
- Target frame-rate
- Whether to stream sound

This UI should be shown to the user every time a stream is initiated.

Dadadah:
This PR adds a screen share settings UI modal that pops up after choosing what window/screen to stream. It allows you to choose a screen share quality. Right now it only supports 720p 30fps, 1080p 30fps and Source 5fps. These values will only show up if the server is configured to allow streams of that size. The `Source@5fps` quality setting will cap the width and height to the configured limit, unless the limit is 0. The modal has a checkbox to not show the modal again. If you check this box, it will save your choice and disable the modal. You can enable it in the settings. You can also change the default screen share resolution chosen in the settings. The `Voice` settings section has been renamed to `Voice & Video`.

This PR also adds a `ButtonGroup` option to the Form2 component.

UI was changed from a select to a button group upon feedback from @infi. Old form2 code was removed and will be put in it's own PR.

~~Requires stoatchat/javascript-client-sdk#149~~
~~Requires #1112~~

This PR is blocking #1070 and #1068

Closes #1069 

## How was this PR tested?

Shared screen from one client to another. Shared multiple screen resolutions and both landscape and portrait screens.

Simulated low resolution cap to ensure only low quality streaming was allowed (720p).

In action:

https://github.com/user-attachments/assets/2efeadc8-b77a-44ef-9a0a-182048eb4354

Settings UI:
image

Available resolutions:
image

</details>

## Known Issues
 - If a Stoat instance has a video resolution limit that is abnormal (ie. not a standard) it may result in undesired results.
 - If a user attempts to start a stream with a weird aspect ratio, that user may get kicked.
   - Unsure on a proper solution for this, if a user starts the stream at a normal aspect ratio and makes it weird they will not get kicked.
 - New user limitations are not respected
   - Unsure if it's possible to get if a user is new currently - may also depend on the server?
     - skill issue?
 - Once a stream is started, it's quality cannot currently be changed.
   - Out of scope for this PR. Could be done separately later.
 - Stream resolution options are not translated.
   - I cannot figure out how to get this to work for the life of me.

## Checklist:

Generative AI was not used in the creation of this PR.


packages/client/components/ui/components/features/voice/callCard/ParticipantTile.tsx
```diff
@@ -72,87 +77,89 @@ export function ParticipantTile(props: TileProps) {
   };
 
   return (
```
Dadadah, line 79: I know the diff looks big here, but I actually just wrapped this in a big show block to disable showing screen shares if they're muted (as is the case when a screen share is getting configured)

Dadadah:
I can't get these strings in the ScreenShareQualities.ts file to translate, it seems like it's executing outside the solid context. I can pass the lingui macro down the chain to the functions, but I'm not sure if there's a better way to do it?

amycatgirl:
IIRC lingui provides a macro called `t` which works like the Trans macro, but for regular strings

Perhaps that's what you are looking for?

(the thing being translated are the full quality names, right?)

Dadadah:
> IIRC lingui provides a macro called `t` which works like the Trans macro, but for regular strings
> 
> Perhaps that's what you are looking for?
> 
> (the thing being translated are the full quality names, right?)

Yes, I'm using `t` but lingui doesn't translate them. I think we have to use the `useLingui` hook, as documented in the docs, but it doesn't work here due to it not having access to the context (I have no idea why that's the case either)

Dadadah:
The solution to this might actually be to move the screenshare qualities into a context. I'm gonna pursue that and see if it turns out alright.

Dadadah:
I can't figure out how to get the translations to work. Rather than hold this critical feature up for it, I just added a TODO mentioning it. This PR should be ready for review.

infi (approved):

mihaicm93 (approved):
Tested on a self-hosted instance worked great

gabrielfordevelopment (approved):
I would separate the resolution and the fps to different buttons/options (so later if we will have more fps options, we don't have to make a huge descartes dropdown/button crowd)
But this really just a minor UX nitpick
Other than that, looks good to me

Dadadah:
Waiting on one more piece of feedback from infi then I'll get this merged.

PR #1131, opened 2026-04-22
feat: Add an error for messages being too long and a countdown

### LINKED ISSUE #1125: Long messages fail to send. (edit: max length is not displayed)
### What happened?

When trying to send a big message it just says "Failed to send" and doesn't say why. It could be an implicit max length or something else entirely, I cannot know. I just happen to have found that textwalls don't send. Might not even be an intentional thing for all I know.

Stoat lacks understandability for new users, many things might feel normal for experienced users but as a newcomer I keep finding unexplained or obscure restrictions that are never explicited and that are left for the user to figure out after trial and error. _Good thing I'm used to puzzles._

Dadadah:
This PR adds a message length ticker that shows when you are 200 characters away from hitting the message limit. It shows red when you are over the limit. It will not allow you to send the message if you are over the limit. It respects the limits in the server side configuration.

Fixes #1125

## Tests
Tested with a 2000 character limit on my self-hosted instance.

[Screencast from 2026-04-23 17-27-19.webm](https://github.com/user-attachments/assets/19996c0e-ba4c-4644-ba02-151bf3f17bb1)

</details>

## Checklist:

- [ ] I have confirmed that any new dependencies are strictly necessary

Generative AI was not used in the writing of this PR.

### DIFF
--- a/packages/client/src/interface/channels/text/Composition.tsx
+++ b/packages/client/src/interface/channels/text/Composition.tsx
@@ -68,12 +68,27 @@ export function MessageComposition(props: Props) {
     return state.draft.getDraft(props.channel.id);
   }
 
+  const messageLength = () => draft().content?.length ?? 0;
+
+  const maxMessageLength = () => {
+    const cl = client();
+    return cl.configured()
+      ? (cl.configuration?.features.limits.default.message_length ?? 2000)
+      : 2000;
+  };
+
+  const isAlmostTooLong = () => messageLength() > maxMessageLength() - 200;
+
   // Whether the send button should be active/clickable
   const canSend = createMemo(() => {
     const draftContent = draft()?.content ?? "";
     const draftFiles = draft()?.files ?? [];
 
-    return draftContent.trim().length > 0 || draftFiles.length > 0;
+    const tooLong = messageLength() > maxMessageLength();
+
+    return (
+      !tooLong && (draftContent.trim().length > 0 || draftFiles.length > 0)
+    );
   });
 
   // TEMP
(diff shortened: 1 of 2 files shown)


gabrielfordevelopment:
I literally had this branch ready, was just waiting for the scrollbar updates to take effect kekw.
Guess I can drop that.

Looks good, but the placement/size could be a question for UI/UX.
You checked the max limit? What if someone pastes in 10 000 long text? Does the number overflow with the chat itself? I guess setting it up to display "too long" once it reaches like `max-limit + 500` characters would improve the UX and fixes potential UI breaks

gabrielfordevelopment:
https://github.com/orgs/stoatchat/discussions/27

gabrielfordevelopment:
Does this apply to Editing a message too? It should

Dadadah:
> Does this apply to Editing a message too? It should

I agree, I'll investigate that tonight.

> Looks good, but the placement/size could be a question for UI/UX. You checked the max limit? What if someone pastes in 10 000 long text? Does the number overflow with the chat itself? I guess setting it up to display "too long" once it reaches like `max-limit + 500` characters would improve the UX and fixes potential UI breaks

I think you would have to reach something like -100,000 or -1,000,000 to reach an overflow. However, I can probably make it say Too Long at like -1000.

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -68,12 +68,27 @@ export function MessageComposition(props: Props) {
     return state.draft.getDraft(props.channel.id);
   }
 
+  const messageLength = () => draft().content?.length ?? 0;
+
+  const maxMessageLength = () => {
+    const cl = client();
+    return cl.configured()
+      ? (cl.configuration?.features.limits.default.message_length ?? 2000)
```
Pecacheu, line 76: I'm getting an error here that limits does not exist on type, maybe my stoat.js is out of date? Git doesn't think there's anything to pull

Dadadah:
> Does this apply to Editing a message too? It should

The edit box doesn't use the same system as the message box and it probably needs to get refactored. I think it might be better to make a git issue to change that separately to use the composition component.

packages/client/src/interface/channels/text/Composition.tsx
Dadadah: This indicates your submodule is out of date, did you try running `git submodule update`?

gabrielfordevelopment:
>  I can probably make it say Too Long at like -1000.

Could you update the screenshot/video pretty please?
Edit: To include the "Too long" message, so we see how it looks on the UI

Dadadah:
> > I can probably make it say Too Long at like -1000.
> 
> Could you update the screenshot/video pretty please? Edit: To include the "Too long" message, so we see how it looks on the UI

I've updated the example to include the "Too Long" message

packages/client/src/interface/channels/text/Composition.tsx
Pecacheu: Just tried it even with `--force` but to no avail. It did change the HEAD of stoat.js but I still get that error that limits doesn't exist, went in there and don't see anything called limits in the schema.ts file either.

gabrielfordevelopment:
> I've updated the example to include the "Too Long" message

I meant 1000 not 10 000 lol.
That (`+1000`) is 150% of the total limit, which should - in my opinion - considered as already "too long"
Also I would change it to: `998... 999... "Too long"` instead of `998... 999...1000... "Too long"` would feel better, so the digits won't jump rapidly within the last few characters.

packages/client/src/interface/channels/text/Composition.tsx
Dadadah: Did you run `mise install:frozen`?

gabrielfordevelopment (approved):

chrishultin (approved):

packages/client/src/interface/channels/text/Composition.tsx
chrishultin: Idea: Perhaps we ought to look into setting up some basic environment/repo setup scripts - something to make sure the appropriate submodules are pulled and up-to-date, for instance.

infi (approved):

packages/client/src/interface/channels/text/Composition.tsx
Pecacheu: > Did you run `mise install:frozen`?

Forgot to reply to this but yep this seems to have been the problem

PR #1138, opened 2026-04-25
fix: call cb on role deletion to redirect to roles list

### LINKED ISSUE #1137: Deleteing a role results in a broken UI
### What happened?

When deleting a role, the settings UI stays on the role page.

### Reproduce
1. Create a server
2. Create a role on the server
3. Delete the role
4. Observe that you are still on the role's page even after confirming delete

### Expected Behavior
After deleting a role, the settings should be redirected to the "Roles" page.

jedbillyb:
The `DeleteRoleModal` was not calling `props.cb()` on successful deletion, 
so the navigation callback passed by `ServerRoleEditor` never fired, 
leaving the user on the deleted role's page.

Fixes #1137

**How was this PR tested?**
- [x] Test A - Created a role, deleted it, confirmed redirect to roles list

### DIFF
--- a/packages/client/components/modal/modals/DeleteRole.tsx
+++ b/packages/client/components/modal/modals/DeleteRole.tsx
@@ -1,8 +1,6 @@
 import { Trans } from "@lingui-solid/solid/macro";
 import { useMutation } from "@tanstack/solid-query";
-
 import { Dialog, DialogProps } from "@revolt/ui";
-
 import { useModals } from "..";
 import { Modals } from "../types";
 
@@ -13,10 +11,10 @@ export function DeleteRoleModal(
   props: DialogProps & Modals & { type: "delete_role" },
 ) {
   const { showError } = useModals();
-
   const deleteRole = useMutation(() => ({
     mutationFn: () => props.role.delete(),
     onError: showError,
+    onSuccess: () => props.cb(),
   }));
 
   return (
--- a/packages/client/components/modal/modals/DeleteRole.tsx
+++ b/packages/client/components/modal/modals/DeleteRole.tsx
@@ -1,4 +1,5 @@
 import { Trans } from "@lingui-solid/solid/macro";
+
 import { useMutation } from "@tanstack/solid-query";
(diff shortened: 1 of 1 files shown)


packages/client/components/modal/modals/DeleteRole.tsx
```diff
@@ -1,8 +1,6 @@
 import { Trans } from "@lingui-solid/solid/macro";
 import { useMutation } from "@tanstack/solid-query";
-
```
Dadadah, line 3: Removing this spacing changes the semantics of the imports. Please revert the imports to their original order.

Dadadah (changes requested):

packages/client/components/modal/modals/DeleteRole.tsx
jedbillyb: Fixed, reverted the import spacing.

Dadadah: You've only reverted one instance; there are still multiple places where your change remove lines on the import order. Please revert all changes to the imports as they are not necessary for this change.

jedbillyb: Fixed.

Dadadah:
Please fix build issues. Also, please DCO your commits. For more information regarding DCO you can see [here](https://github.com/stoatchat/for-web/pull/1138/checks?check_run_id=73129377140).

packages/client/components/modal/modals/DeleteRole.tsx
```diff
@@ -10,13 +10,14 @@ import { Modals } from "../types";
  * Modal to delete a role
  */
 export function DeleteRoleModal(
-  props: DialogProps & Modals & { type: "delete_role" },
+  props: DialogProps & Modals & { type:
```
infi, line 13: Is there a reason for this change? That looks like it's leaving the file with invalid syntax to me.

infi (changes requested):

Dadadah:
closing as stale

PR #1161, opened 2026-05-12
feat: add ability to copy links from messages

Dadadah:
Add ability to right click copy links from messages.
Rename `Copy link` to `Copy message link` in the context menu to avoid confusion.
Add a message context to allow easier access to message data in markdown (this process is a pain, a context makes it easy.)

Do note that the context menu cannot use the context. It exists outside of it in the floating manager, therefore it must still have the `message` prop.

## How was this PR tested?
Right clicked a bunch of messages with and without links to ensure context menu was as expected.

image
</details>

## Checklist:

Generative AI was not used in the writing of this PR.

### DIFF
--- a/packages/client/components/app/interface/channels/text/Message.tsx
+++ b/packages/client/components/app/interface/channels/text/Message.tsx
@@ -84,14 +84,21 @@ interface Props {
   isLink?: boolean;
 }
 
-const messageContext = createContext<MessageInterface | undefined>(undefined);
+interface MessageContextShape {
+  message?: MessageInterface;
+}
+
+const messageContext = createContext<MessageContextShape>({});
 
 function MessageContext(props: {
   children: JSX.Element;
   message: MessageInterface;
 }) {
+  // eslint-disable-next-line solid/reactivity
+  const contextShape = { message: props.message };
+
   return (
-    <messageContext.Provider value={props.message}>
+    <messageContext.Provider value={contextShape}>
       {props.children}
     </messageContext.Provider>
   );
--- a/packages/client/components/app/interface/channels/text/Message.tsx
+++ b/packages/client/components/app/interface/channels/text/Message.tsx
@@ -1,4 +1,14 @@
-import { For, Match, Show, Switch, createSignal, onMount } from "solid-js";
+import {
+  For,
+  JSX,
+  Match,
+  Show,
+  Switch,
+  createContext,
+  createSignal,
+  onMount,
+  useContext,
+} from "solid-js";
 
 import { useLingui } from "@lingui-solid/solid/macro";
 import {
@@ -74,6 +84,21 @@ interface Props {
   isLink?: boolean;
 }
 
+const messageContext = createContext<MessageInterface | undefined>(undefined);
+
+function MessageContext(props: {
+  children: JSX.Element;
+  message: MessageInterface;
+}) {
+  return (
+    <messageContext.Provider value={props.message}>
+      {props.children}
+    </messageContext.Provider>
+  );
+}
+
+export const useMessage = () => useContext(messageContext);
+
 /**
  * Render a Message with or without a tail
  */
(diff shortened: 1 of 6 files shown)


Arsabutispik (approved):

Pecacheu:
Works good! My only question is if untrusted links should still allow copying them... but I think allowing it is probably fine since right-clicking to open a context menu then pressing "Copy Link" is a much more intentional action than just clicking on it... and that feature is mainly meant to stop you from clicking on suspicious links absent mindedly without thinking.

Pecacheu (approved):

Pecacheu (changes requested):
Oh hold on, just realized there's something I forgot to test- LinkComponents can also appear in other markdown contexts besides messages, such as a bio- in which cause useMessage presumably is undefined.

Just gave that a test and the result looks pretty weird. Imo the context menu should just be disabled in this case.

Dadadah:
> Oh hold on, just realized there's something I forgot to test- LinkComponents can also appear in other markdown contexts besides messages, such as a bio- in which cause useMessage presumably is undefined.

Yeah that was it, I fixed the issue. Good catch.

Pecacheu (approved):
Forgot to approve this, works now.

Only issue is I get a reactivity warning here, so that should either be fixed or get an eslint-disable


Edit: Second thing, I realized that the useMessage() context would come in handy for something I'm passing as a prop to components of the message in #835, but in order to pass that value as well useMessage will have to provide an object so we can extend it with other properties it can provide.

I.e. `const message = useMessage();` would become `const { message } = useMessage();`

packages/client/components/app/interface/channels/text/Message.tsx
```diff
@@ -65,6 +75,21 @@ interface Props {
...
+function MessageContext(props: {
+  children: JSX.Element;
+  message: MessageInterface;
+}) {
+  return (
+    <messageContext.Provider value={props.message}>
+      {props.children}
+    </messageContext.Provider>
+  );
+}
```
Pecacheu:
```suggestion
interface MessageCtx {
  message?: MessageInterface;
}

const messageContext = createContext<MessageCtx>(null as never as MessageCtx);

function MessageContext(props: {
  children: JSX.Element;
  message: MessageInterface;
}) {
  /* eslint-disable solid/reactivity */
  return (
    <messageContext.Provider
      value={{
        message: props.message,
      }}
    >
      {props.children}
    </messageContext.Provider>
  );
}
```

packages/client/components/markdown/plugins/anchors.tsx
```diff
@@ -241,11 +244,28 @@ export function RenderAnchor(
...
 
 function LinkComponent(
-  props: { disabled?: boolean } & JSX.AnchorHTMLAttributes<HTMLAnchorElement>,
+  props: {
+    disabled?: boolean;
+    dest?: string;
+  } & JSX.AnchorHTMLAttributes<HTMLAnchorElement>,
 ) {
-  const [localProps, remoteProps] = splitProps(props, ["disabled"]);
+  const message = useMessage();
```
Pecacheu:
```suggestion
  const { message } = useMessage();
```

packages/client/components/ui/components/features/messaging/elements/Container.tsx
```diff
@@ -306,10 +304,11 @@ const CompactInfo = styled(Row, {
  */
 export function MessageContainer(props: Props) {
   const { t } = useLingui();
+  const message = useMessage();
```
Pecacheu:
```suggestion
  const { message } = useMessage();
```

packages/client/components/ui/components/features/messaging/elements/MessageToolbar.tsx
```diff
@@ -17,13 +17,13 @@ import MdEmojiEmotions from "@material-design-icons/svg/outlined/emoji_emotions.
...
 
-import { startsWithPackPUA } from "@revolt/markdown/emoji/UnicodeEmoji";
 import { CompositionMediaPicker } from "../composition";
 
-export function MessageToolbar(props: { message?: Message }) {
+export function MessageToolbar() {
   const user = useUser();
   const state = useState();
   const { openModal } = useModals();
+  const message = useMessage();
```
Pecacheu:
```suggestion
  const { message } = useMessage();
```

Pecacheu (changes requested):
Suggestion for allowing expandability as mentioned above

packages/client/components/app/interface/channels/text/Message.tsx
```diff
@@ -65,6 +75,21 @@ interface Props {
...
+
+export const useMessage = () => useContext(messageContext);
```
Pecacheu, line 107:
```suggestion
export const useMessage = () => useContext(messageContext) ?? {};
```

Oops, above change causes an undefined error without this

Pecacheu (changes requested):

Pecacheu (approved):

PR #1166, opened 2026-05-16
feat: slowmode in text channels

Arsabutispik:
Adds slowmode to TextChannels with the ability to set them, set permissions for roles to be able to bypass them too.

Currently there are a few issues;
- User can still send messages while active slowmode, even if the server forces that to fail (I can work on this before the UI/UX reviewer looks at the code)
- The persistence doesn't apply trough different clients or websites. It would be only persistent in a single client (I am thinking about would using user settings work here?)

Fixes # (no issue was created for this)

## How was this PR tested?

- [x] Set slowmode in a text channel, people with perms were not affected by it.
- [x] An account that didn't have perms sent messages, first 5 seconds of slowmode, was able to send after 5 seconds.
- [x] Set a higher amount (30 seconds for example) reset site changed active slowmode to see that user still only had to go trough 30 seconds of slowmode.

image
image
image
image

</details>

## Checklist:

none used


Arsabutispik:
The current way of keeping the slowmodes "persistent" is not pretty. I will work on a backend feature to be able to use it in here

packages/client/components/app/interface/settings/channel/Overview.tsx
```diff
@@ -27,13 +36,15 @@ export default function ChannelOverview(props: ChannelSettingsProps) {
     icon: createFormControl<string | File[] | null>(
       props.channel.animatedIconURL,
     ),
+    slowmode: createFormControl(String(props.channel.slowmode ?? 0)),
```
Dadadah: Instead of casting this to a string and then back to a number, store it as a number and convert it to a string when operating with it in the floating select.

Dadadah: Please add a type to the createFormControl, like line 36.

Arsabutispik:
This time to keep it persistent it uses Userslowmode event sent from backend

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -52,6 +53,70 @@ export function MessageComposition(props: Props) {
   const client = useClient();
   const { openModal } = useModals();
 
+  const [now, setNow] = createSignal(Date.now());
+
+  const currentSlowmode = createMemo(() =>
```
Dadadah: I don't think this needs to be a memo, it should only change once, when the slowmode expires. It can be an accessor instead.

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -52,6 +53,70 @@ export function MessageComposition(props: Props) {
...
+  );
+
+  createEffect(() => {
+    if (cooldownRemaining() > 0) {
+      const timer = setInterval(() => setNow(Date.now()), 1000);
+      onCleanup(() => clearInterval(timer));
+    }
+  });
+
+  const isSlowmodeExempt = createMemo(() => {
```
Dadadah: I don't think this needs to be a memo. It can be an accessor.

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -52,6 +53,70 @@ export function MessageComposition(props: Props) {
   const client = useClient();
   const { openModal } = useModals();
 
+  const [now, setNow] = createSignal(Date.now());
```
Dadadah: Please investigate the `@solid-primitives/date` package, specifically the `createCountdown` hook. I think you can rewrite a big chunk of this code with that hook and it should make things a lot easier as it will have reactivity built in and you don't need to do these shenanigans of creating a weird reactive timer with the interval below. If you need help implementing this feel free to reach out over DMs.

https://primitives.solidjs.community/package/date/#createCountdown

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -344,6 +417,32 @@ export function MessageComposition(props: Props) {
           );
         }}
       </For>
+      <Show when={props.channel.slowmode}>
+        <div
```
Dadadah: Please create styled components for this div, the div on line 432, and the span on like 436.

Dadadah (changes requested):

Arsabutispik:
Some sort of change happened between my last changes and now I can't get it to working at all, I don't see any useful errors or my attempts of debugging, not even console.debug worked, has been thawed and left me in an extremely furious state. And since my exams are closing in (then I will have a job) I won't be working on this feature any more.

Arsabutispik:
After coming back to this with a cool head, I realised the issue was the production, not me.

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -52,6 +56,91 @@ export function MessageComposition(props: Props) {
...
+    slowmodeKey();
+
+    const entry = currentSlowmode();
+    if (!entry) return null;
+
+    const receivedAt = entry.receivedAt ?? Date.now();
+    const targetTs = receivedAt + entry.retry_after * 1000;
+
+    return createCountdownFromNow(targetTs);
+  });
```
Dadadah: I think I understand what you're doing here, you're writing to the slowmodeKey every time you receive a `userSlowmodes` event so you can recalculate the countdown. It does look like this should work, but I think it's more convoluted than it has to be. I believe you can just listen to the userSlowmodes store directly. Lastly, try to return `undefined` instead of `null` unless you must as it allows some syntactic shortucts that null doesn't. Give this suggestion a try and see if the linter complains about it being reactive, and if it seems to work:

```suggestion
  const currentSlowmode = (): UserSlowmodes => {
    return client().userSlowmodes.get(props.channel.id);
  }

  const countdownForEntry = createMemo(() => {
    const entry = currentSlowmode();
    if (!entry) return;

    const receivedAt = entry.receivedAt ?? Date.now();
    const targetTs = receivedAt + entry.retry_after * 1000;

    return createCountdownFromNow(targetTs);
  });
```

The client userSlowmodes object is a ReactiveMap, so it should have reactivity built in, making it not necessary to wait for the userSlowmodes event from the client to recalculate it, as the client will already update the userSlowmodes map.

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -52,6 +56,91 @@ export function MessageComposition(props: Props) {
...
+
+  function isSlowmodeExempt() {
+    return props.channel.havePermission("BypassSlowmode");
+  }
```
Dadadah: While this works, accessors are usually defined in this codebase with arrow functions `() => {}`

```suggestion
  const isSlowmodeExempt = (): boolean => {
    return props.channel.havePermission("BypassSlowmode");
  }
```

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -52,6 +56,91 @@ export function MessageComposition(props: Props) {
...
+
+  const cooldownRemaining = createMemo(() => {
+    if (!props.channel.slowmode || isSlowmodeExempt()) return 0;
+
+    const cd = countdownForEntry();
+    if (!cd) return 0;
+
+    // createCountdownFromNow returns [store, metadata], access store at [0]
+    const store = cd[0];
```
Dadadah: You can use an array initializer here and it should make this line more clear:

```suggestion
    const [ store ] = cd;
```

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -52,6 +56,91 @@ export function MessageComposition(props: Props) {
...
+    const h =
+      typeof store.hours === "function" ? store.hours() : (store.hours ?? 0);
+    const m =
+      typeof store.minutes === "function"
+        ? store.minutes()
+        : (store.minutes ?? 0);
+    const s =
+      typeof store.seconds === "function"
+        ? store.seconds()
+        : (store.seconds ?? 0);
```
Dadadah: Do you need to check if these are functions? I think the countdown object is always just numbers.

packages/client/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "@codemirror/autocomplete": "^6.19.0",
+    "@solid-primitives/date": "^2.1.2",
```
Dadadah: Little confused as to why this isn't alphabetized, you used `pnpm i @solid-primatives/date` right? Could you alphabetize this?

packages/client/src/interface/channels/text/Composition.tsx
```diff
@@ -435,3 +546,26 @@ export function MessageComposition(props: Props) {
...
+    justifyContent: "flex-end",
+    padding: "0 12px 6px 0",
+  },
+});
+
+const SlowmodeRow = styled("div", {
+  base: {
+    display: "flex",
+    alignItems: "center",
+    gap: "4px",
```
Dadadah: Please use the built in css gaps

```suggestion
    gap: "var(--gap-sm)",
```

Dadadah (changes requested):

Dadadah (approved):

Dadadah:
A pr was just merged that changed how the floating select works in forms (using Form2.Select.) Could you update your PR to use this?

PR #1170, opened 2026-05-17
fix: remove member option was not showing correctly

amycatgirl:
Fixes the remove member from group option because it was not showing up correctly due to faulty logic. Also moves the menu option into the correct position.

## How was this PR tested?

- [x] Was able to see the option when user is the owner of the channel and is not targeting self or the owner.
N/a
</details>

## Checklist:

- [ ] I have made corresponding changes to the documentation if applicable (N/a)
- [ ] I have confirmed that any new dependencies are strictly necessary (N/a)
- [ ] I have written tests for new code (if applicable) (N/a)

none this time around

### DIFF
--- a/packages/client/components/app/menus/UserContextMenu.tsx
+++ b/packages/client/components/app/menus/UserContextMenu.tsx
@@ -293,8 +293,9 @@ export function UserContextMenu(props: {
     return (
       props.channel?.type === "Group" &&
       !props.user.self &&
-      props.channel.havePermission("ManageChannel") &&
-      !props.channel.owner?.self
+      !(props.channel.owner?.id !== props.user.id) &&
+      (props.channel.havePermission("ManageChannel") ||
+        props.channel.owner?.self)
     );
   }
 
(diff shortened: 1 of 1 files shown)


packages/client/components/app/menus/UserContextMenu.tsx
```diff
@@ -293,8 +293,9 @@ export function UserContextMenu(props: {
     return (
       props.channel?.type === "Group" &&
       !props.user.self &&
-      props.channel.havePermission("ManageChannel") &&
-      !props.channel.owner?.self
+      !(props.channel.owner?.id !== props.user.id) &&
```
Dadadah: I tested this in my self host and I believe this line is inverted. It should be `props.channel.owner?.id !== props.user.id &&` instead of `!(props.channel.owner?.id !== props.user.id) &&`

packages/client/components/app/menus/UserContextMenu.tsx
Dadadah: Your last PR added this divider, can you remove it in this PR?

Dadadah (changes requested):

Dadadah:
There is currently a bug in stoat.js that makes the client think all group members have the `ManageChannel` permission. This probably won't full work as expected until that is fixed.

packages/client/components/app/menus/UserContextMenu.tsx
amycatgirl: should be resolved in f800a7f

packages/client/components/app/menus/UserContextMenu.tsx
amycatgirl: Removed in f800a7f; I think that was because i was merging changes with upstream and merged both hunks instead of keeping the remote one.

amycatgirl:
> There is currently a bug in stoat.js that makes the client think all group members have the ManageChannel permission. This probably won't full work as expected until that is fixed.

ah, that's a bit inconvenient

Dadadah:
I agree, but it's better we get this working, and report the bug on stoat.js.

Dadadah (approved):

PR #1195, opened 2026-05-23
fix: Config fixes for TSConfig root dir & VSCode search ignores

Pecacheu:
Just two small config improvements made due to annoyance dealing with these issues lol.

- Add the ignores from gitignore to vscode search exclusion to speed up search (Exclusions can be manually overriden in the search panel if need be)
- Add tsconfigRootDir to eslint to fix multiple TS config candidates warning in vscode/eslint that pops up sometimes

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,5 +20,11 @@
   "editor.insertSpaces": true,
   "editor.tabSize": 2,
   // Nix environment settings
-  "nixEnvSelector.nixFile": "${workspaceFolder}/default.nix"
+  "nixEnvSelector.nixFile": "${workspaceFolder}/default.nix",
+  "search.exclude": {
+    "**/coverage": true,
+    "**/dist": true,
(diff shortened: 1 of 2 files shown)


Dadadah:
I think the tsconfigrootdir change absolutely nuked the linter.

Pecacheu:
Oof, slight oversight, lint and format check fails due to it using that config for all the submodules too. This can probably be fixed by not linting submodules (because why would you lint submodules, they're read-only)

Dadadah:
I think you need to set the `tsconfigRootDir` to just `import.meta.dirname` and not add `packages/client` to it. The linter error when it does show up suggests that.

Edit:

image

Dadadah:
We should probably add this to stoat.js too because it will keep complaining about it there.

Pecacheu:
Woo it works now

Dadadah (approved):

PR #1196, opened 2026-05-23
feat: Device resize hooks (prep for Mobile/Tablet UI)

Pecacheu:
In preparation for Mobile/Tablet UI, we need to add a centralized place for reactive hooks for layout changes and feature detection.

For best performance for the resize hooks, I went with a hybrid approach that uses native CSS media queries and forwards them to reactive variables. From my testing this performs better than a resize observer (though either is obviously way better than brute-force polling).

This will become a prerequisite for <https://github.com/stoatchat/for-web/pull/835> which will use the hooks here.

- `main` 
  - \#1196 :point\_left:

### DIFF
--- a/packages/client/components/common/Device.tsx
+++ b/packages/client/components/common/Device.tsx
@@ -46,7 +46,7 @@ export class Device {
 
     this.pMedia = matchMedia(`(max-width: ${this.phoneMaxWidth})`);
     this.tMedia = matchMedia(`(max-width: ${this.tabletMaxWidth})`);
-    (this.pMedia.onchange = this.tMedia.onchange = this.onLayout)();
+    (this.pMedia.onchange = this.tMedia.onchange = this.onLayout.bind(this))();
   }
 
   onLayout() {
--- a/packages/client/components/common/Device.tsx
+++ b/packages/client/components/common/Device.tsx
@@ -1,4 +1,5 @@
 import {
+  Accessor,
   createContext,
   createSignal,
   JSX,
@@ -25,13 +26,13 @@ export class Device {
   readonly tabletMaxWidth = style.getPropertyValue("--tablet-max-width");
 
   /** Layout type based on viewport size */
-  readonly layout;
+  readonly layout: Accessor<Layout>;
 
   /** Mobile device detection based on User Agent.
 
    * **Warning:** Don't use unless absolutely necessary.
    * Granular feature-detection is preferred when possible. */
-  readonly isMobile;
+  readonly isMobile: boolean;
 
   private pMedia;
   private tMedia;
--- a/packages/client/components/common/Device.tsx
+++ b/packages/client/components/common/Device.tsx
@@ -11,11 +11,7 @@ import { isMobileBrowser } from "@livekit/components-core";
 
 const style = getComputedStyle(document.body);
 
-export enum Layout {
-  DESKTOP = 1,
-  PHONE,
-  TABLET,
-}
+export type Layout = "desktop" | "tablet" | "phone";
 
 /** Device type and compatibility info */
 export class Device {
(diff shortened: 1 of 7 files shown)


Pecacheu:
Okay, dunno what's going on with GitHub's diff but it is incorrect. This didn't change and I just diffed it manually against the version currently in base/main.


The only change to this file is adding DeviceContext.

infi (commented):
No visible UI changes, UI/UX approval not required

packages/client/components/common/Device.tsx
```diff
@@ -0,0 +1,78 @@
...
+/** Device type and compatibility info */
+export class Device {
+  /** Max width for phone layout */
+  readonly phoneMaxWidth = style.getPropertyValue("--phone-max-width");
+
+  /** Max width for tablet layout */
+  readonly tabletMaxWidth = style.getPropertyValue("--tablet-max-width");
+
+  /** Layout type based on viewport size */
+  readonly layout;
```
Dadadah: Please add type annotations to variables not initialized inline.

packages/client/components/common/Device.tsx
```diff
@@ -0,0 +1,78 @@
...
+  constructor() {
+    this.isMobile = isMobileBrowser();
+
+    const [lo, setLo] = createSignal<Layout>(Layout.DESKTOP);
+    this.layout = lo;
+    this.setLayout = setLo;
+
+    this.pMedia = matchMedia(`(max-width: ${this.phoneMaxWidth})`);
+    this.tMedia = matchMedia(`(max-width: ${this.tabletMaxWidth})`);
+    (this.pMedia.onchange = this.tMedia.onchange = this.onLayout.bind(this))();
```
Dadadah, line 46: Please use `addEventListener("change" . . . )` and `removeEventListener("change" . . .)` instead of setting `onchange`, as is recommended by MDN.

packages/client/components/ui/styles.css
```diff
@@ -6,8 +6,11 @@
 @import "@fontsource/inter/800";
 @import "@fontsource/jetbrains-mono";
 
-* {
-  box-sizing: border-box;
+* { box-sizing: border-box; }
+
+:root {
+  --phone-max-width: 600px;
+  --tablet-max-width: 900px;
```
Dadadah: Change tablet to `960px` as the team decided to match m3ui's breakpoints which are 600px and 960px

Dadadah (changes requested):
I like this hook. We should probably add some documentation on the developer documentation about how and when to use it.

packages/client/components/ui/styles.css
Pecacheu: Will do. I was just going for rounded numbers haha

packages/client/components/common/Device.tsx
Pecacheu: I mean, normally I agree it's a good idea to always use those but the reason for that is that if anyone else tries to set the same listener on the same element, it overwrites it if you use the `onX` property. But in this case, the media objects are totally private and internal to the class, so no one else can access them.

packages/client/components/common/Device.tsx
Pecacheu: Will do. I think eslint only let me get away with it because those are set in the constructor

packages/client/components/common/Device.tsx
```diff
@@ -0,0 +1,79 @@
...
+  JSX,
+  onCleanup,
+  useContext,
+} from "solid-js";
+
+import { isMobileBrowser } from "@livekit/components-core";
+
+const style = getComputedStyle(document.body);
+
+export enum Layout {
```
Dadadah: Sorry didn't see this on the first pass. Please use a string union type instead of an enum as we don't really use enums in this codebase. It'd be `export type Layout = "desktop" | "tablet" | "mobile";`.

Pecacheu: What can I say I'm a C developer I hate strings and prefer enums lol
Didn't know there was a preference in this codebase tho, I can change that pretty easy (well other than that I now have to change it in 835 upstream too)

Dadadah (approved):

PR #1199, opened 2026-05-24
feat: Add shorthand names for some emojis

Pecacheu:
This was actually split out from #1003 but I forgot to create a PR for it.

This simply adds a few emoji shorthand names (same as what Discord has where some emojis can be triggered by multiple names). If you think of any I missed, lmk. Combined with 1003 it allows you to quickly type emojis by starting with an opening `:`, followed by the shorthand name, then end with another `:` (or use tab) and it appears instantly.

Given how often I see people ending messages on Stoat with `:tm:` and expecting it to work but instead nothing happens, clearly this is needed lol. Heck I even saw Insert do it once :P

Built on:
- #1202


Dadadah (changes requested):
This file is automatically generated by the script at `lifecycle/generateEmojiMapping.ts` which means your changes here will be overwritten in the next emoji update.

Instead, you should add the shortcuts to the generateEmojiMapping.ts file, and rerun it (using Deno) and commit the results.

Pecacheu:
Ah, good catch, I didn't realize that file is generated. I think I can solve that by adding a dict with the alternate names to generateEmojiMapping.ts.

PR #1202, opened 2026-05-24
chore: Update mise tools and add generateemoji task

Dadadah:
Update mise tools to get us pnpm 11

Also fixes prettier bug by forcing a single prettier version

Fix mise build:check

Fix a bunch of type errors

- `main` 
  - \#1202 :point\_left:


Pecacheu:
Awesome, it works!

Kind of annoying that it doesn't type-hint the deno imports or methods tho. I'm not familiar with deno so not sure how to fix that, maybe you need a matching `@types` package for TypeScript?

Pecacheu (approved):

PR #1217, opened 2026-05-30
feat: Multi-user account switching

Pecacheu:
Split off from and baseline for multi-instance switching in:
- #999

## How was this PR tested?
At least a dozen furries daily-driving PR 999 for months :3

This has been briefly tested to ensure that extraction from the above didn't break it.

Back button appears on auth page only when adding an additional account, not on first login. Matches styling of the light/dark mode button.

</details>

## Checklist:

80% Raichu powered, 20% V8 Energy drink at 2 AM powered

### DIFF
--- a/packages/client/components/auth/src/AuthPage.tsx
+++ b/packages/client/components/auth/src/AuthPage.tsx
@@ -1,14 +1,17 @@
 import { BiLogosGithub } from "solid-icons/bi";
-import { JSX } from "solid-js";
+import { JSX, Show } from "solid-js";
 
 import { Trans } from "@lingui/solid/macro";
 import { styled } from "styled-system/jsx";
 
 import { Titlebar } from "@revolt/app/interface/desktop/Titlebar";
+import { useClientLifecycle } from "@revolt/client";
+import { State } from "@revolt/client/Controller";
 import { useState } from "@revolt/state";
 import { IconButton, iconSize } from "@revolt/ui";
 
 import MdDarkMode from "@material-design-icons/svg/filled/dark_mode.svg?component-solid";
+import MdArrowBack from "@material-design-icons/svg/outlined/arrow_back.svg?component-solid";
 
 import background from "./background.jpg";
 import { FlowBase } from "./flows/Flow";
@@ -30,9 +33,8 @@ const Base = styled("div", {
     flexDirection: "column",
     justifyContent: "space-between",
 
-    mdDown: {
-      padding: "30px 20px",
-    },
+    _tablet: { padding: "30px 20px" },
+    _phone: { padding: "10px 5px" },
   },
 });
 
@@ -75,7 +77,6 @@ const NavItems = styled("div", {
     gap: "10px",
     display: "flex",
     alignItems: "center",
-
     fontSize: "0.9em",
   },
   variants: {
@@ -86,9 +87,8 @@ const NavItems = styled("div", {
         },
       },
       stack: {
-        md: {
-          flexDirection: "column",
-        },
+        flexDirection: "column",
+        _tablet: { flexDirection: "row" },
       },
     },
   },
@@ -109,13 +109,14 @@ const LinkWithIcon = styled("a", {
  */
 const Bullet = styled("div", {
   base: {
+    display: "none",
     height: "5px",
     width: "5px",
     background: "grey",
     borderRadius: "50%",
 
-    md: {
-      display: "none",
+    _tablet: {
+      display: "block",
     },
   },
 });
@@ -124,7 +125,8 @@ const Bullet = styled("div", {
  * Authentication page
  */
 export function AuthPage(props: { children: JSX.Element }) {
-  const state = useState();
+  const state = useState(),
+    ctrl = useClientLifecycle();
 
   return (
     <Root>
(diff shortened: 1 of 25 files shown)


chrishultin (approved):

chrishultin (approved):

packages/client/components/auth/src/AuthPage.tsx
```diff
@@ -132,7 +135,17 @@ export function AuthPage(props: { children: JSX.Element }) {
         css={{ scrollbar: "hidden" }}
       >
         <Nav>
-          <div />
+          <Show
+            when={
+              ctrl.lifecycle.state() === State.Ready &&
+              state.auth.hasMultiSession()
+            }
+            fallback={<div />}
```
Dadadah, line 144: Is this fallback needed? I don't see a reason to show an empty node, it can just not have a fallback.

packages/client/components/client/Controller.ts
```diff
@@ -480,32 +474,33 @@ export default class ClientController {
...
-  isLoggedIn() {
-    return this.isLoggedInState();
-  }
-
   isError() {
     return this.lifecycle.state() === State.Error;
   }
 
+  loginCached(unhold = false, cached = unhold) {
+    console.log("AUTH loginCached", unhold, cached);
```
Dadadah: Rogue log

Dadadah: line 503 as well

Dadadah: 620, 628

Dadadah: 643

packages/client/components/client/Controller.ts
```diff
@@ -597,16 +593,54 @@ export default class ClientController {
...
-  logout() {
-    this.state.settings.resetNotificationsState();
-    killServiceWorkerSubscription(this.getCurrentClient());
-    this.state.auth.removeSession();
+  #cacheUserInfo() {
+    const user = this.lifecycle.client.user;
+    if (user) this.state.auth.cacheUserInfo(user);
+  }
+
+  #swapping = false;
```
Dadadah: Define this near the top of the class like other class variables

packages/client/components/auth/src/flows/Flow.tsx
```diff
@@ -20,14 +20,15 @@ export const FlowBase = styled("div", {
     background: "var(--md-sys-color-surface-container)",
     color: "var(--md-sys-color-on-surface)",
     width: "360px",
-    maxWidth: "360px",
-    maxHeight: "600px",
+    maxWidth: "100%",
```
Dadadah: Why are you changing the max width here, is this related to this pr?

packages/client/components/state/stores/Auth.ts
```diff
@@ -1,13 +1,18 @@
...
+import { User } from "stoat.js";
 
 import { AbstractStore } from ".";
+import { State } from "..";
 
 export type Session = {
   _id: string;
   token: string;
   userId: string;
+  host?: string;
```
Dadadah: Host? I don't think we need that in this pr, is this a remnant of the self-host pr?

Dadadah (changes requested):
Not bad, I would appreciate if you could go in to detail about how the state store works now, perhaps we should update the developer documentation?

This PR is large, are you sure there aren't any more relevant screenshots or videos for ui/ux review?

packages/client/components/client/Controller.ts
Pecacheu: Yup just forgot to remove the logging used for testing (tho it was highly useful at the time for ensuring the states transition correctly and such lol)

packages/client/components/auth/src/flows/Flow.tsx
Pecacheu: Oop this is for one thing I forgot to add a screenshot for- Back button added to login page so that you can go back if you change your mind about signing in a second account. It only appears if you have at least one account already and not at first sign-in, ofc. But it required some minor CSS tweaks to make it appear in the right spot

packages/client/components/state/stores/Auth.ts
Pecacheu: Ah you are right. As is this will just always be null since it is never set anywhere

packages/client/components/auth/src/AuthPage.tsx
Pecacheu: Idk man the empty div was already there before lol

Will have to double check if removing it breaks anything.

Pecacheu: Okay yea figured it out- It's a placeholder to ensure the button positioning on the top bar is correct, as the light/dark mode button is floated right using flexbox space-between. A bit of an unusual choice, but the fact they were doing it that way actually made it super easy to add a second button on the left that appears sometimes.

Pecacheu:
> This PR is large, are you sure there aren't any more relevant screenshots or videos for ui/ux review?

Just added one more screenshot- But aside from that, this PR is more backend changes than UI so that's that's changed UI-wise. Eg. several of the changed JSX files was just to pass the onClose listener to the add account button so that it closes the settings UI when pressed.

> Not bad, I would appreciate if you could go in to detail about how the state store works now, perhaps we should update the developer documentation?

Sure, updated documentation for the State/Store changes might make sense. Long story short, now instead of having only one database (currently named "localforage" in main, as that's what the library defaults to), we treat the old one as the 'global' DB that holds any store with the global flag set to true- namely the auth store at the moment. All other stores are instead loaded from and written to the appropriate store with the name `instanceHost@userId` - Since only default instance is supported at the moment, the former is blank, i.e. it is just `@userId`.

For forward compatibility, I made the multi-instance PR behave like this too, so that none of your settings for existing main-instance logins are lost in the void. In other words, he default instance (as set by CONFIGURATION) is always replaced with a blank string (even if you explicitly type in `stoat.chat` in the instance field) so that there is no instance prefix. And your first logged-in account stores everything in "localforage" instead, again so that existing settings are not lost.

One thing I didn't yet implement that should probably be added is garbage collection for unused DBs, i.e. after logout of a secondary account the data just sits there. The databases known to localforage should be double-checked against the list of saved sessions in the global auth store.

PR #1279, opened 2026-06-19
fix: Add base path to service worker registration check

Dadadah:
Fixes issue in production where the service worker registration cannot be fetched.

Thanks @amycatgirl for help finding the root cause of this issue.

- `main` 
  - \#1279 :point\_left:

### DIFF
--- a/packages/client/components/client/NotificationsController.ts
+++ b/packages/client/components/client/NotificationsController.ts
@@ -136,7 +136,9 @@ async function setUpServiceWorkerSubscription(client: Client) {
     throw "Client not configured";
   }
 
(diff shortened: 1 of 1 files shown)


amycatgirl:
Looks fine at first glance, I'll be deploying this in a test environment to check if it fixes the issue.

amycatgirl (approved):
It works 🎉

Tested it by building the app using `PROD="true" mise build:prod` and using a basic http server that served the dist folder inside of the `/app/` route.

Push notifications could be enabled and disabled correctly.

infi (approved):
Ship it

PR #1301, opened 2026-06-25
fix: Mobile-friendly changes for emoji picker

Pecacheu:
Split off from #835

- The GIF/emoji picker effectively "takes the place" of the touch-keyboard until you explicitly open it by tapping the search bar, like on Discord.
  - Don't autofocus search input in GIF/emoji picker on mobile because it causes unwanted touch keyboard deployment
  - Don't refocus the message box element still hidden behind picker after tapping an Emoji, for same reason
- Shrink emoji picker when screen is too small (reactively calculates row count depending on picker width)
- Prevent picker height from exceeding screen height
- Fix overflow hidden clipping emoji autocorrect dropdown when editing a message

* `main` 
  - \#1301 :point\_left:

### DIFF
--- a/packages/client/components/ui/components/features/messaging/composition/picker/CompositionMediaPicker.tsx
+++ b/packages/client/components/ui/components/features/messaging/composition/picker/CompositionMediaPicker.tsx
@@ -131,6 +131,7 @@ function Picker(
     if (rect.right > innerWidth || rect.bottom > innerHeight) setFixed(true);
   }
   onMount(() => {
+    (document.activeElement as HTMLElement)?.blur(); //Hide keyboard
     addEventListener("mousedown", onMouseDown);
     addEventListener("resize", onResize);
     setTimeout(onResize, 1);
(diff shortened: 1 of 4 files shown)


Dadadah:
I am running into an issue with this in dev - when opening the picker while the keyboard is open it closes my keyboard but then doesn't move to the bottom.


https://github.com/user-attachments/assets/7f37e161-6527-46d7-90c2-7b767bd12d42

Pecacheu:
> I am running into an issue with this in dev - when opening the picker while the keyboard is open it closes my keyboard but then doesn't move to the bottom.
> 
>  screen-20260809-233101.mp4

Ah I think I see what's happening, it's trying to avoid overflowing the top of the screen and thus sticks to the top, but that's overriding it trying to stick to the bottom. My phone screen is taller cause I have a fat ass Galaxy Ultra phone so this doesn't happen to me lol

I'll have to see how I can fix that order of priorities

Dadadah (approved):

PR #1302, opened 2026-06-25
fix(messages): scroll edited message into view when entering edit state

### LINKED ISSUE #1299: Pressing the up arrow key ⬆️ to edit should scroll the message into view
### What happened?

In a channel, when you press the up arrow key in a channel where you have previously sent a message and that message is loaded by the app, you start editing this message, which is a lot more convenient than having to right click or even move your mouse at all to fix up your message.

However, when the message is scrolled out of view, the app does not scroll it into view, leaving it the responsibility of the user to find the message.

Instead a message should be scrolled into view automatically when it enters the editing state.

Antoniodeoliveirasegura:
Closes #1299. Pressing ↑ now scrolls the message being edited into view when it’s off-screen, reusing the existing `Messages` lookup pattern with `block: "nearest"` so already-visible messages don’t move.

### DIFF
--- a/packages/client/components/app/interface/channels/text/Messages.tsx
+++ b/packages/client/components/app/interface/channels/text/Messages.tsx
@@ -884,6 +884,36 @@ export function Messages(props: Props) {
   );
 
   /**
+   * Scroll the message being edited into view when it enters the editing state.
+   * Editing can be triggered by the up-arrow keybind on a message that has been
+   * scrolled off-screen, in which case it should be brought back into view.
+   */
+  createEffect(
+    on(
+      () =>
+        typeof state.draft.editingMessageId === "string"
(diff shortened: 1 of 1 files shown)


Dadadah:
Please sign your commits. You can see more information about that [here](https://github.com/stoatchat/for-web/pull/1302/checks?check_run_id=83337504236).

Please fill out the full PR template.

Antoniodeoliveirasegura:
Thanks for the review, @Dadadah and apologies. I missed that the contributing guidelines ask contributors not to open LLM-generated PRs, and this one was AI-generated, so it doesn't meet that policy. Closing it; sorry for the noise. I'll leave #1299 for someone else to pick up.

PR #1303, opened 2026-06-25
fix: cache material symbols to prevent them from breaking on bad conn

Dadadah:
Add the material symbols to the service worker cache so that mobile connections don't result in the fonts messing up

Also adds a preload to the index.html via a vite plugin, meaning you shouldn't ever see the fonts messing up again (but it mnight take a bit longer to load the app sometimes)

- `main` 
  - \#1303 :point\_left:

### DIFF
--- a/packages/client/fontpreload.plugin.ts
+++ b/packages/client/fontpreload.plugin.ts
@@ -1,0 +1,51 @@
+import type { HtmlTagDescriptor, Plugin } from "vite";
+
+// This plugin was based on the following Stack Overflow answer
+// https://stackoverflow.com/a/79671988/5709936
+
+export function addFontPreload(): Plugin {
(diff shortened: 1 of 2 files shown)


Pecacheu:
I think we can take this out of draft, I've got in on my test instance and it seems to be working fine.

Dadadah:
Jenn brought up making the fonts a preload which will also prevent the loading in font issue, but I haven't gotten around to it.

Pecacheu:
If it's preloaded and not cached, won't that slow down loading of the app every time? Or does marking it as preload automatically cause the browser to cache it?

Dadadah:
it would slow it down, however, I mean that I'd like to include that with this pr to get the best of both worlds.

infi (approved):

PR #1310, opened 2026-06-26
fix: collected messages array staying on preemt breaking chat

Dadadah:
Another regression from the collected messages change. If the messages were preempted the chat you're in would break forever (until it reloaded for some reason) because it would be collecting infinitely.

This fixes that hopefully by setting the `collectedMessages` array to undefined on preempt.

- `main` 
  - \#1310 :point\_left:

### DIFF
--- a/packages/client/components/app/interface/channels/text/Messages.tsx
+++ b/packages/client/components/app/interface/channels/text/Messages.tsx
@@ -234,7 +234,10 @@ export function Messages(props: Props) {
       }
 
       // Cancel if we've been pre-empted
-      if (preempted()) return;
+      if (preempted()) {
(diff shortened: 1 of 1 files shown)


Pecacheu:
Haven't looked at this area of the code in detail but is there any reason you don't need to add this to the couple of other places where a return happens after a preempt() check, like line 331 for instance?

Dadadah:
> Haven't looked at this area of the code in detail but is there any reason you don't need to add this to the couple of other places where a return happens after a preempt() check, like line 331 for instance?

The other preempt calls don't use the collectedMessages array.

Pecacheu (approved):

PR #1312, opened 2026-06-28
feat: Add backend server emoji limit

Dadadah:
Add the backend limit for server emojis.

Useful for self host.

Partially addresses #1133

## How was this PR tested?

Set emoji limit to 1000 in my self host

<details><summary><h2>Screenshots & Screencasts (if appropriate)</h2></summary>
image

</details>

## Checklist:

- [x] I have carefully read [the contributing guidelines](https://developers.stoat.chat/developing/contrib/)
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation if applicable
- [x] I have no unrelated changes in the PR
- [x] I have confirmed that any new dependencies are strictly necessary
- [ ] I have written tests for new code (if applicable)
- [x] I have followed naming conventions/patterns in the surrounding code

## Please declare, if any, LLM usage involved in creating this PR

None

- `main` 
  - \#1312 :point\_left:

PR #1315, opened 2026-06-29
feat: Instance context to enable routes and reactivity w/ consistent interface

Pecacheu:
Requires:

- <https://github.com/stoatchat/javascript-client-sdk/pull/172>

Perquisite for:

- \#999
- \#1217

Supersedes:

- \#1309
- \#1312

Adds the Instance context for accessing instance-specific configurations and detecting the API endpoint from the current instance. Also adds an easy-to-use memo to the ClientController that simplifies limit checks and safely returns defaults if client is unconfigured, and drops some now-unused CONFIGURATION env vars.

*Note: This PR **does not** actually enable instance switching, nor does it add/change any UI. But it is the groundwork for doing so, adding the necessary reactivity, and enabling the route handling for alternate instances.*

Edit: Messed around with the commit history a bit to make sure Gtoasted gets credited for their initial contributions to this.

- `main` 
  - \#1315 :point\_left:

### DIFF
--- a/packages/client/components/app/menus/UserContextMenu.tsx
+++ b/packages/client/components/app/menus/UserContextMenu.tsx
@@ -61,7 +61,7 @@ export function UserContextMenu(props: {
    * Open direct message channel
    */
   function openDm() {
-    props.user.openDM().then((channel) => navigate(`/channel/${channel.id}`));
+    props.user.openDM().then((channel) => navigate(channel.path));
     props.onClose?.();
   }
 
--- a/packages/client/components/instance/Instance.ts
+++ b/packages/client/components/instance/Instance.ts
@@ -1,0 +1,62 @@
+export default class Instance {
+  readonly isStoat: boolean;
+  readonly apiUrl: string;
+  readonly wsUrl: string;
+  readonly mediaUrl: string;
+  readonly proxyUrl: string;
+  readonly gifboxUrl: string;
+  readonly hcaptcha_sitekey: string;
+  readonly maxEmoji: number;
+  readonly enableVideo: boolean;
+  readonly hostname: string | undefined;
+
+  // Not implemented, but should be fine for now
+  // readonly maxReplies: number;
+  // readonly maxAttachments: number;
+  // readonly maxFileSize: number;
+  // DEVELOPMENT_SESSION_ID
+  // DEVELOPMENT_TOKEN
+  // DEVELOPMENT_USER_ID
+
+  constructor(
+    apiUrl: string,
+    wsUrl: string,
+    mediaUrl: string,
+    proxyUrl: string,
+    gifboxUrl: string,
+    hcaptcha_sitekey: string,
+    maxEmoji: number,
+    enableVideo: boolean,
+    hostname?: string,
+  ) {
+    this.isStoat = [
+      // historically...
+      "https://api.revolt.chat",
+      "https://beta.revolt.chat/api",
+      "https://revolt.chat/api",
+      // ... and now:
+      "https://stoat.chat/api",
+    ].includes(apiUrl);
+    this.apiUrl = apiUrl;
+    this.wsUrl = wsUrl;
+    this.mediaUrl = mediaUrl;
+    this.proxyUrl = proxyUrl;
+    this.gifboxUrl = gifboxUrl;
+    this.hcaptcha_sitekey = hcaptcha_sitekey;
+    this.maxEmoji = maxEmoji;
+    this.enableVideo = enableVideo;
+    this.hostname = hostname;
+  }
+
+  /**
+   * Prepends the given url with this instances' base path.
+   * @param url - The url to link to.
+   */
+  href(url: string) {
+    if (!this.hostname) {
+      return url;
+    }
+
+    return `/instance/${this.hostname}${url}`;
+  }
+}
--- a/packages/client/components/instance/Instance.ts
+++ b/packages/client/components/instance/Instance.ts
@@ -1,62 +1,110 @@
+import { Navigator } from "@solidjs/router";
+import { Accessor, createMemo } from "solid-js";
+
+import { CONFIGURATION } from "@revolt/common";
+import { AppConfig, STOAT_HOST } from "@revolt/common/lib/env";
+import { Client, UserLimits } from "stoat.js";
+
+import { DefaultHost, StoatOrigin } from ".";
+
+const R_RelPath = /^\/i\/[^/]+/;
+
 export default class Instance {
+  readonly host?: string;
+  readonly origin: string;
   readonly isStoat: boolean;
+  readonly #base;
+  readonly #nav;
+
   readonly apiUrl: string;
   readonly wsUrl: string;
   readonly mediaUrl: string;
   readonly proxyUrl: string;
   readonly gifboxUrl: string;
-  readonly hcaptcha_sitekey: string;
-  readonly maxEmoji: number;
-  readonly enableVideo: boolean;
-  readonly hostname: string | undefined;
 
-  // Not implemented, but should be fine for now
-  // readonly maxReplies: number;
-  // readonly maxAttachments: number;
-  // readonly maxFileSize: number;
-  // DEVELOPMENT_SESSION_ID
-  // DEVELOPMENT_TOKEN
-  // DEVELOPMENT_USER_ID
+  #firstInit = true;
+  /** Note: This is slightly risky- If you import with const braces, eg. `const { client } = useInstance()`,
+   * it won't be reactive. `useClient()` is preferred to avoid ambiguity. */
+  readonly client!: Client;
+  readonly config;
 
-  constructor(
-    apiUrl: string,
-    wsUrl: string,
-    mediaUrl: string,
-    proxyUrl: string,
-    gifboxUrl: string,
-    hcaptcha_sitekey: string,
-    maxEmoji: number,
-    enableVideo: boolean,
-    hostname?: string,
-  ) {
-    this.isStoat = [
-      // historically...
-      "https://api.revolt.chat",
-      "https://beta.revolt.chat/api",
-      "https://revolt.chat/api",
-      // ... and now:
-      "https://stoat.chat/api",
-    ].includes(apiUrl);
-    this.apiUrl = apiUrl;
-    this.wsUrl = wsUrl;
-    this.mediaUrl = mediaUrl;
-    this.proxyUrl = proxyUrl;
-    this.gifboxUrl = gifboxUrl;
-    this.hcaptcha_sitekey = hcaptcha_sitekey;
-    this.maxEmoji = maxEmoji;
-    this.enableVideo = enableVideo;
-    this.hostname = hostname;
+  //Instead of these, use the reactive limits accessor below when possible
+  readonly globalLimits;
+  readonly baseLimits;
+
+  /** Current enforced limits based on user type */
+  readonly limits: Accessor<UserLimits>;
+
+  constructor(appCfg: AppConfig, cli: Client, host: string, nav: Navigator) {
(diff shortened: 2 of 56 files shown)


packages/client/src/index.tsx
```diff
@@ -116,80 +117,84 @@ function BotRedirect() {
...
-              <Route path="/*" component={ServerHome} />
-            </Route>
-            <Route path="/channel/:channel/*" component={ChannelPage} />
-            <Route path="/*" component={HomePage} />
+      <SnackbarProvider controller={snackbarCtrl}>
+        <Router>
+          {/* <Route path="/i/:host" component={InstanceContext}>
+            {routes()}
+          </Route> */}
+          <Route path="/" component={InstanceContext}>
```
Dadadah: It looks like in this change you have removed the root node, ie. the Route with `root` prop. What does this prop do, can we do without it, and can you explain your reasoning for removing it?

packages/client/src/interface/navigation/channels/HomeSidebar.tsx
```diff
@@ -92,7 +94,9 @@ export const HomeSidebar = (props: Props) => {
             href="/friends"
             size="normal"
             icon={<Symbol>group</Symbol>}
-            attention={location.pathname === "/friends" ? "selected" : "normal"}
+            attention={
+              location.pathname.endsWith("/friends") ? "selected" : "normal"
+            }
```
Dadadah, line 104: Could you, instead of using endsWith here, use a helper function in the instance context that determines if you're at that location inside the instance context?

packages/client/components/state/index.tsx
```diff
@@ -218,7 +218,7 @@ const stateContext = createContext<State>(null! as State);
 /**
  * Mount state context
  */
-export function StateContext(props: { children: JSX.Element }) {
+export function StateContext(props: { children?: JSX.Element }) {
```
Dadadah: Why would children be undefined here?

packages/client/components/instance/Instance.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    //Endpoints
+    this.apiUrl = appCfg.api;
+    this.wsUrl = apiCfg.ws;
+    this.mediaUrl = apiCfg.features.autumn.url;
+    this.proxyUrl = apiCfg.features.january.url;
+    this.gifboxUrl = "https://api.gifbox.me"; //TODO Gifbox URL from backend
+
+    //Features
+    this.captchaKey = CONFIGURATION.HCAPTCHA_SITEKEY; //TODO Detect from API
```
Dadadah: Captcha key is available under configuration.features.captcha.key

packages/client/components/instance/Instance.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.isStoat = host === STOAT_HOST;
+    this.#base = this.isStoat ? "" : `/i/${host}`;
+    this.#nav = nav;
+  }
+
+  /** Prepend a relative path with instance base URL
+   * @param pathOnly Get only the path component and not URL
+   * @param base Defaults to the base path of this instance
+   */
+  href = (path: string, pathOnly?: boolean, base?: string) =>
```
Dadadah, line 80: I like this function. Can we get one for media urls as well? This should make handling media urls elsewhere (like UserProfileEditor.tsx:133) easier.

packages/client/components/instance/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+    try {
+      const appCfg: AppConfig = host
+        ? await (await fetch(`https://${host}/.well-known/stoat`)).json()
+        : { api: CONFIGURATION.DEFAULT_API_URL };
+
+      const cli = _newClient(appCfg.api);
+      let instSet = false;
+
+      //TODO Need to catch API connection error somehow and redirect to onError
+      createEffect(() => {
```
Dadadah: Can you confirm (or implement) #1327 with this?

Dadadah (changes requested):
This looks fine for the most part. I'm a little worried about the complexity the instance introduces, if this is something you're worried about as well perhaps we should open a dialog to talk about how to reduce complexity of this system.

Do you need to update any documentation regarding the client lifecycle? Perhaps double check for that. I'm gonna want more eyes on this if possible.

packages/client/components/app/menus/UserContextMenu.tsx
```diff
@@ -61,7 +61,7 @@ export function UserContextMenu(props: {
    * Open direct message channel
    */
   function openDm() {
-    props.user.openDM().then((channel) => navigate(channel.url));
+    props.user.openDM().then((channel) => navigate(channel.path));
```
Dadadah, line 64: This change doesn't belong in this PR. Please remove it.

packages/client/components/ui/components/features/profiles/ProfileActions.tsx
```diff
@@ -31,7 +31,7 @@ export function ProfileActions(props: {
    * Open direct message channel
    */
   function openDm() {
-    props.user.openDM().then((channel) => navigate(channel.url));
+    props.user.openDM().then((channel) => navigate(channel.path));
```
Dadadah, line 34: This change doesn't belong in this PR. Please remove it.

Dadadah (changes requested):
Sorry forgot a few things.

packages/client/components/app/menus/UserContextMenu.tsx
Pecacheu: This change is required because the instance stripping doesn't work using the full URL parameter, only path. I can open an issue for this bug if needed, but it only shows up when there's a prefix on the URL, which previously didn't happen in this specific area until now where all URLs have the potential to have an instance prefix.

packages/client/components/instance/index.tsx
Pecacheu: Ooh those new screens look nice. I'll rebase and give that a try

packages/client/components/instance/Instance.ts
Pecacheu: Good to know, will remove that from hard-coding

packages/client/components/state/index.tsx
Pecacheu: This was throwing a TS compilation error without it, I'm not sure why as it clearly worked before without it. But the error even happened on build in the GitHub actions, so it wasn't just my system having the wrong TS config or something

packages/client/components/ui/components/features/profiles/ProfileActions.tsx
Pecacheu: Same explaination as above

packages/client/src/index.tsx
Pecacheu: Not removed, just moved. It's now at line 150. Basically the order of the hierarchy is the same as before, except it's all inside of the instance context, except for the Device and Snackbar contexts, which remain outside it. Device context for obvious reasons as that doesn't change based on what instance you're in, and Snackbars are needed to report some errors eg. when connecting to instance.

packages/client/src/interface/navigation/channels/HomeSidebar.tsx
Pecacheu: Good idea, that should allow it to still match against the full path but ignoring the instance prefix

Pecacheu: Oh, silly me I already had a method for this and I didn't use it here lol, it's instance.relPath

packages/client/components/app/menus/UserContextMenu.tsx
Pecacheu: Edit: Actually your PR at #1334 may have fixed this issue already. Though before that change the profile Message button was working fine for me UNTIL I add any instance prefix, so idk why that user had that bug to report in the first place lol

packages/client/components/instance/Instance.ts
Pecacheu: Hmm, gave this a try but it feels a little, idk... Captain obvious-y? Here is the method:
```ts
mediaHref = (path: string) => this.mediaUrl + path;
```

It didn't really make the usage more consise, eg:
``newBannerUrl = `${instance.mediaUrl}/backgrounds/${changes.profile.background}`;``
to
``newBannerUrl = instance.mediaHref(`/backgrounds/${changes.profile.background}`);``

I think the other href method makes more sense cause it's doing a bit more than just concatenating the two.

Dadadah: ya on second thought you might be right here. You can ignore this comment.

packages/client/components/instance/index.tsx
Pecacheu: Okay, tested and working with the new loading screen! Had to move LoadingScreen.tsx to somewhere more standard and re-export it like other components in order to import it without a nightmare fuel `../../../../../` situation lol

Dadadah (approved):
LGTM. Lets merge this after 0.13.0 so we can get pronouns out without worrying about this breaking anything.

Pecacheu:
Awesome, sounds good to me

packages/client/components/ui/components/design/LoadingScreen.tsx
```diff
@@ -96,12 +86,12 @@ export function LoadingScreen() {
...
 
   onMount(() => {
-    if (!isEligibleOrigin()) return;
-
     const controller = new AbortController();
 
     // Check the status API for an ongoing incident.
     const statusTimer = setTimeout(async () => {
+      //TODO Fetch status from current instance backend instead of only stoat.chat
+      if (!instance.isStoat) return;
```
chrishultin: instance might be undefined - can we handle that?

packages/client/components/instance/index.tsx
```diff
@@ -0,0 +1,125 @@
...
```
chrishultin: Can we add a timeout path if the configuration never loads?

packages/client/components/instance/Instance.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.wsUrl = apiCfg.ws;
+    this.mediaUrl = apiCfg.features.autumn.url;
+    this.proxyUrl = apiCfg.features.january.url;
+    this.gifboxUrl = "https://api.gifbox.me"; //TODO Gifbox URL from backend
+
+    //Features
+    this.config = apiCfg;
+    this.globalLimits = apiCfg.features.limits.global;
+    this.baseLimits = apiCfg.features.limits.new_user;
+    this.limits = createMemo(() => this.client.limits ?? this.baseLimits);
```
chrishultin: newClient() still reassigns this.client - //@ts-expect-error readonly

I think this means that client gets overridden on every logout.

chrishultin (dismissed):

packages/client/components/ui/components/design/LoadingScreen.tsx
Pecacheu: Fixed, will push soon

packages/client/components/instance/Instance.ts
Pecacheu: Yes this is intended behavior, Client is replaced when the user changes (this is also how it was done before on logout/login), but I've set it to readonly just so that no one outside the class can overwrite it by accident, it only gets overwritten internally. I'll edit the comments to make it more clear

packages/client/components/instance/index.tsx
Pecacheu: Sorry, this comment is outdated. Wasn't handled at the time, but the timeout case is now handled by the try/catch along with LoadingScreen. First you'll get a "This is taking longer than usual" warning, then finally you'll get a "Couldn't fetch Stoat configuration" error snackbar.

PR #1323, opened 2026-07-03
fix: add automatic date/time format option for locale-native defaults

### LINKED ISSUE #629: Locales that doesn't fit the patterns of date and time have odd results
### What happened?

In the menu for languages, you can choose three formats for date: "DD/MM/YYYY", "MM/DD/YYYY", "YYYY-MM-DD" as well as two formats for time: "HH:MM", "H:MM AA".

The issue is that not many locales will fit into these and will results in odd behaviour. For example, in German, the date-format is `dd.MM.yy` and in Spanish (Spain) the time-format is `H:mm`

On the web, changing the language will reset the selections for date and time.
On desktop, it will preserve the selection if you have made a change, and this is locked forever.
On Android, there's no such option and the app has built-in formats per locale.

The problem on the web is when you select a format, you can't go back to the locale's default. So having an option for both date and time that represents the default/automatic format for the selected locale, would be the best.

On the web, selecting a locale like German, will correctly display "DD.MM.YYYY", and every format is forced to leading zeros and 4 digit years, except Bulgarian, which is displayed as "D.MM.YYYY", most likely due to an error since the format is set to `d.MM.yy 'г'.`.

On the web, selecting a locale using specifically `HH:mm` will correctly show it as "24 hours", but selecting any other locale, regardless if it's 12 or 24 hours, such as Bulgarian or Spanish using `H:mm` or Indonesian using `HH.mm`, will instead display "12 hours", but it will correctly display it correctly as 24 hours in the locale's format. Oddly this also happens to Azerbaijani despite using `HH:mm`.

Both of these cases would be solved having the default/automatic option. This does mean that for some locales, default/automatic will be the same as one of the overrides, but this is fine. Both the web and desktop should keep the selected override and not reset back to default, since the user can now choose if they want the default or not.

Kameti77:
This PR adds an **Automatic** option to both the date and time format pickers in Language settings. Previously, the only choices were fixed presets (DD/MM/YYYY, MM/DD/YYYY, YYYY-MM-DD for date and HH:mm, h:mm A for time). Many locales use formats that don't match any of these. For example, German uses DD.MM.YYYY, Spanish uses H:mm, and Indonesian uses HH.mm, so users on those locales had no option that correctly reflected their locale's native format, and there was no way to revert to the locale default once an override had been picked.

The Automatic option uses the locale's native dayjs format tokens (L for date, LT for time), so it always displays the correct format for whatever language is selected. Switching languages updates the Automatic preview immediately. Explicit overrides are preserved across language switches and can be reverted to Automatic at any time.

**Changes made:**
- `Language.tsx` — Added Automatic as the first option in both pickers. The preview is computed via `createMemo` tracking a new `nativeLocaleFormats` signal so it updates reactively when the locale changes. Selecting Automatic calls `clearDateFormat()`/`clearTimeFormat()` on the store.
- `dayjs.tsx` — Added a `nativeLocaleFormats` signal that captures the locale's original L and LT formats before any user overrides are applied, so the Automatic preview is never affected by a previously picked override.
- `Locale.ts` — Added a public `getOptions()` accessor (the existing `get()` is protected). Fixed `clearDateFormat()` and `clearTimeFormat()` to call `loadTimeLocale()` directly so the locale reloads even when the language itself hasn't changed.
- `Time.tsx` — Changed the `date` format case to use dayjs's `L` token (locale-aware) instead of the hardcoded `DD/MM/YYYY`. Added a separate `dateNormal` case that keeps the hardcoded `DD/MM/YYYY` for the Traditional option preview. Added `createEffect` so the component updates immediately when the locale signal changes rather than waiting for the next `setInterval` tick.

Fixes #629

## How was this PR tested?

- [ ] Switched to German — Date format shows "Automatic" with preview `DD.MM.YYYY` (dots)
- [ ] Switched to Spanish — Time format shows "Automatic" with preview `H:mm` (no AM/PM, no leading zero)
- [ ] Switched to Azerbaijani — Time format shows "Automatic" with correct 24-hour format
- [ ] Switched to English (Traditional) — Automatic date preview shows `DD/MM/YYYY`, time shows `h:mm A`
- [ ] Switched to English (Simplified) — Automatic date preview shows `MM/DD/YYYY`
- [ ] Picked an explicit override (`DD/MM/YYYY`), then switched languages — override was preserved, did not reset
- [ ] Selected Automatic after an explicit override — preview reverted to locale-native format immediately
- [ ] Refreshed the page with an explicit override — override persisted correctly
- [ ] Refreshed the page with Automatic selected — still showed Automatic with correct locale preview
- [ ] Verified Traditional (DD/MM/YYYY), American (MM/DD/YYYY), and ISO Standard previews always show their fixed formats regardless of locale

Screenshot 2026-07-02 215356

Screenshot 2026-07-02 215431

</details>

## Checklist:

- [x ] I have carefully read [the contributing guidelines](https://developers.stoat.chat/developing/contrib/)
- [x ] I have made corresponding changes to the documentation if applicable

I used Claude Code (Anthropic) as a pair programming assistant throughout this PR. It helped me navigate the codebase, understand the Solid.js reactivity model, and reason through the timing issues with async locale loading. All code was reviewed and understood before committing.


Kameti77:
Hi @stoatchat/for-web-maintainers, this is my first contribution. I've fixed #629 by adding an Automatic option to the date/time format pickers so locales like German and Spanish correctly show their native formats. Would appreciate a review when you have time! Looking forward to your feedback.

Dadadah:
Please review our contribution guidelines. We do not accept PRs made with LLMs.

PR #1327, opened 2026-07-05
feat: helpful loading screen

infi:
- **feat: loading screen with status querying**
- **feat: also show troubleshooting advice after 10s**

## How was this PR tested?

Proprietary E.Y.E.S. technology

<h2>Screenshots & Screencasts (if appropriate)</h2>

image
image
image

</details>

- `main` 
  - \#1327 :point\_left:

### DIFF
--- a/packages/client/src/LoadingScreen.tsx
+++ b/packages/client/src/LoadingScreen.tsx
@@ -1,5 +1,5 @@
 import { Trans, useLingui } from "@lingui-solid/solid/macro";
-import { Show, createSignal, onCleanup, onMount } from "solid-js";
+import { JSX, Show, createSignal, onCleanup, onMount } from "solid-js";
 import { styled } from "styled-system/jsx";
 
 import { Button, CircularProgress, Symbol, Text } from "@revolt/ui";
@@ -10,6 +10,12 @@ import { Button, CircularProgress, Symbol, Text } from "@revolt/ui";
 const STATUS_PAGE_URL = "https://status.stoat.chat";
 
 /**
+ * Connection troubleshooting knowledge base article
+ */
+const TROUBLESHOOTING_URL =
+  "https://support.stoat.chat/kb/troubleshooting/connection-issues";
+
+/**
  * Status API queried when the client is slow to connect
  */
 const STATUS_API_URL = "https://stoat.chat/-/status";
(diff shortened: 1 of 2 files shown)


chrishultin (approved):

packages/client/src/LoadingScreen.tsx
```diff
@@ -0,0 +1,280 @@
...
+};
+
+/**
+ * Whether the current origin is eligible to probe the status API
+ */
+const isEligibleOrigin = () => {
+  const { hostname } = window.location;
+  return (
+    hostname === "localhost" || // (for testing)
+    hostname.endsWith(".stoat.chat") ||
```
chrishultin, line 54: Nit: Could be reduced to just `hostname.endsWith("stoat.chat")`

chrishultin: Meaning you can remove the `||` and following line

Arsabutispik: Wouldn't it be better to just see which API we are trying to connect to? If it's stoat we attempt to get service status

packages/client/src/LoadingScreen.tsx
```diff
@@ -0,0 +1,280 @@
...
+    hostname === "stoat.chat"
+  );
+};
```
Dadadah, line 57: This is duplicate code, as this check already exists in the android-nag code. Can we move this check into a centralized location? I think it would fit well in the same location as the device hook (in it's own file), which is located in `packages/client/components/common`, or optionally in the lib folder: `packages/client/components/common/lib`.

Dadadah (changes requested):

packages/client/src/LoadingScreen.tsx
infi: > Wouldn't it be better to just see which API we are trying to connect to? If it's stoat we attempt to get service status

No, the API is cors locked to those domains (and a few more that are owned by RPL)

infi: > Nit: Could be reduced to just `hostname.endsWith("stoat.chat")`

That would catch `this-is-not-stoat.chat`

Dadadah (approved):

PR #1364, opened 2026-07-21
fix: PWA mobile rotation-lock fix & system bar theme color

Pecacheu:
Fixes a recently reported issue where the app continues to rotate even if the OS-level Rotation Lock is enabled. This is because setting orientation to `any` in Vite overrode the system setting.

Also fixes the system bar color being locked to the color of the default theme even if you change theme colors, by reactively matching the meta theme color to the current UI state.

## How was this PR tested?

PWA build in dev and production mode and tested on Android in Brave and Firefox.

## Screenshots

Open/closed slide drawer:

 

Landscape:

URL bar / PWA notification shade during load:

 

## Checklist:

Electric mice Pokemon are not defined as valid end users in ChatGPT's ToS.

- `main` 
  - \#1364 :point\_left:

### DIFF
--- a/packages/client/components/ui/themes/LoadTheme.tsx
+++ b/packages/client/components/ui/themes/LoadTheme.tsx
@@ -1,5 +1,6 @@
-import { createEffect } from "solid-js";
+import { createEffect, createMemo } from "solid-js";
 
+import ClientController, { State } from "@revolt/client/Controller";
 import { useState } from "@revolt/state";
 
 import {
@@ -7,6 +8,7 @@ import {
   createMduiColourTriplets,
   createStoatWebVariables,
 } from ".";
+import { SlideState } from "../components/navigation/SlideDrawer";
 import { Masks } from "./Masks";
 import { FONTS, MONOSPACE_FONTS } from "./fonts";
 import { legacyThemeUnsetShim } from "./legacyThemeGeneratorCode";
@@ -14,17 +16,17 @@ import { legacyThemeUnsetShim } from "./legacyThemeGeneratorCode";
 /**
  * Component for loading theme variables into root
  */
-export function LoadTheme() {
+export function LoadTheme(props: { cliCtx?: ClientController }) {
   const state = useState();
 
-  createEffect(() => {
+  const getCssProps = createMemo(() => {
     const activeTheme = state.theme.activeTheme;
 
     // load fonts
     FONTS[state.theme.interfaceFont].load();
     MONOSPACE_FONTS[state.theme.monospaceFont].load();
 
-    for (const [key, value] of Object.entries({
+    const cssProps = {
       // create unset variables to indicate where colours need replacing
       ...Object.keys(legacyThemeUnsetShim().colours).reduce(
         (d, k) => ({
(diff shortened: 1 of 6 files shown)


packages/client/src/index.tsx
```diff
@@ -138,7 +144,7 @@ function MountContext(props: { children?: JSX.Element }) {
           </ClientContext>
         </ModalContext>
       </KeybindContext>
-      <LoadTheme />
+      <LoadTheme cliCtx={cliCtx()} />
```
Dadadah: This is fishy. It looks like you're doing a two directional context which just seems like a poor use of context. What is referencing the LoadTheme that makes it where it can't be inside the client context? I would rather refactor this than do a bidirectional context.

Pecacheu: It was just to keep it so that the theme applies to the app loading screen, but I suppose that's not strictly necessary.

packages/client/components/ui/themes/LoadTheme.tsx
```diff
@@ -41,9 +45,38 @@ export function LoadTheme() {
       ...createMaterialColourVariables(activeTheme, "--md-sys-color-"),
       // mount --mdui-color triplet variables
       ...createMduiColourTriplets(activeTheme, "--mdui-color-"),
-    })) {
+    };
+
+    for (const [key, value] of Object.entries(cssProps))
       document.body.style.setProperty(key, value);
-    }
```
Dadadah, line 46: Could you move the side effects out of this memo? Ie. the document.body.style.setProperty calls and the font loading stuff. I think they can go into the effect below since that effect relies on getCssProps.

Dadadah: the effect below may be called too much, maybe it should go in another effect that is only dependent on getCssProps

Pecacheu: Ok side effects should be moved out of the memo. It needs to be in it's own createEffect though because the other one has several other dependencies that update it too.

Not 100% sure what you mean by the other effect "may be called too much", but maybe you meant that the bannerShown check would make it trigger every time the lifecycle state changes? I moved that to a memo which should fix that, I think.

Dadadah: that was an incomplete thought my bad, I meant it was called too much to put specifically the loading fonts and document style property sets in.

packages/client/components/ui/themes/LoadTheme.tsx
```diff
@@ -16,15 +19,25 @@ import { legacyThemeUnsetShim } from "./legacyThemeGeneratorCode";
...
+      State.Offline,
+    ].includes(lifecycle.state()),
+  );
+
+  const getCssProps = createMemo(() => {
     const activeTheme = state.theme.activeTheme;
 
     // load fonts
     FONTS[state.theme.interfaceFont].load();
     MONOSPACE_FONTS[state.theme.monospaceFont].load();
```
Dadadah: Should the font loading also be put into the effect below?

Not sure what this is doing, but I'm assuming it's loading the fonts for the theme and that'd probably be a side effect as well.

Pecacheu: Hmm that I'm not sure since I didn't write that code either lol

I just did a quick check through the CSS props that get calculated there to check if they are only color related or if there are some font ones, and it looks like there are some font ones being computed. So that font loading may be related/have to go before calculating the CSS vars or else it might use the old font in it's calculation(?)

Pecacheu: Hmm actually never mind because those load() functions are async and not awaited, so the calculation clearly doesn't need to go first (unless we just found another race condition lol)

So I think it can be moved, but I'll test changing fonts to be sure

Dadadah (approved):

PR #1274, opened 2026-06-13
style: use macos native traffic lights

elaraspace:
The macOS app uses its own controls in the title bar. Because many macOS apps uses native "trafic lights" in the top-left of the window, it can possibly be seen inconsistent.

This PR changes title bar UI and these changes only apply in macOS.

Because both `for-web` and `for-desktop` need to be updated, **you have to merge both requests at the same time**. (https://github.com/stoatchat/for-web/pull/1274, https://github.com/stoatchat/for-desktop/pull/220)

Thanks.

### Original version:
Screenshot 2026-06-14 at 6 10 04 AM

Screenshot 2026-06-14 at 6 16 02 AM

### Updated version:
Screenshot 2026-06-14 at 6 10 25 AM

Screenshot 2026-06-14 at 6 15 40 AM

PR #1367, opened 2026-07-22
fix: remove marginInlineStart(70px) when opened in a macOS browser

elaraspace:
I recently created a pull request #1274 to make the macOS app use native traffic lights UI. But it ended up giving unnecessary space for browsers too. This pull request fixes it.

## How was this PR tested?
Yes. I'm attaching two screenshots for comparison.

#### Before
Screenshot 2026-07-23 at 3 26 53 AM

#### After
Screenshot 2026-07-23 at 3 23 34 AM

### DIFF
--- a/packages/client/components/app/interface/desktop/Titlebar.tsx
+++ b/packages/client/components/app/interface/desktop/Titlebar.tsx
@@ -18,7 +18,7 @@ import Wordmark from "../../../../public/assets/web/wordmark.svg?component-solid
 import { pendingUpdate } from "../../../../src/serviceWorkerInterface";
 
 const isMacOS = navigator.platform.startsWith("Mac");
-const isNative = Boolean(window.native);
+const isNative = !!window.native;
 
 export function Titlebar() {
   const [isMaximised, setIsMaximised] = createSignal(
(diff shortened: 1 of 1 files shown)


packages/client/components/app/interface/desktop/Titlebar.tsx
```diff
@@ -18,10 +18,11 @@ import Wordmark from "../../../../public/assets/web/wordmark.svg?component-solid
 import { pendingUpdate } from "../../../../src/serviceWorkerInterface";
 
 const isMacOS = navigator.platform.startsWith("Mac");
+const isNative = Boolean(window.native);
```
Dadadah: This doesn't follow the code styling of the rest of the app. Please instead use `!!window.native`.

Dadadah (changes requested):

packages/client/components/app/interface/desktop/Titlebar.tsx
elaraspace: I pushed the change. Thanks for the feedback!

Dadadah (approved):

PR #1375, opened 2026-07-23
fix: Fix keyboard flicker via VirtualKeyboard API

Pecacheu:
Thanks to the **POWAH** of the new [VirtualKeyboard API](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard), we can finally fix the annoying little flicker and momentary frame-rate stutter that happens whenever you open the virtual keyboard on mobile!

This happens because the viewport is being resized by the browser as the keyboard opens, which takes it a bit to process, whereas a native app doesn't actually resize the viewport- the keyboard overlays the content, but UI elements move to adjust to its size and position. This updates the app to use a similar strategy on supported devices, falling back to the current approach otherwise.

There should (hopefully) be no other visual changes to take screenshots of.

## How was this PR tested?

With Android fone. iOS not supported yet but allegedly coming soon in a future Safari release.

## Checklist:

Sorry, the dog ate my LLM 🙀

- `main` 
  - \#1375 :point\_left:

### DIFF
--- a/packages/client/components/auth/src/AuthPage.tsx
+++ b/packages/client/components/auth/src/AuthPage.tsx
@@ -26,13 +26,6 @@ const Base = styled("div", {
     userSelect: "none",
     overflowY: "scroll",
 
-    color: "var(--md-sys-color-on-surface)",
-    background: "var(--md-sys-color-surface)",
-    // background: `var(--url)`,
-    // backgroundPosition: "center",
-    // backgroundRepeat: "no-repeat",
-    // backgroundSize: "cover",
-
     display: "flex",
     flexDirection: "column",
     justifyContent: "space-between",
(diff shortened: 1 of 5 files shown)


Dadadah (approved):
LGTM! I'm going to test this in my self host to make sure the PWA works as expected, then we can merge this for 0.14.0.

Dadadah (commented):
I don't think this is working in my self host

Pecacheu:
> I don't think this is working in my self host

That's weird, there's not that much to not work. The PWA detection might not be triggering somehow? I'd try using devtools on the PWA and testing if this is true:

```js
isPWA =
    //Safari
    (window.navigator.standalone ||
    //Other
    window.matchMedia("(display-mode: standalone)").matches;
```

As well as if `navigator.virtualKeyboard` is defined.

PR #1380, opened 2026-07-24
feat: Guide users to install PWA

Pecacheu:
This is a rebase of #1077 on latest main + dropping the message.po files, since @Misofist didn't get around to it for a while. I also dropped the PWA SW update change, that should probably be in its own PR since it requires its own testing to ensure it actually works.

On mobile browsers, the "Install" menu appears in settings.

On browsers with support for `beforeinstallprompt`, an easy install button is present.

Pressing it displays the native browser install prompt.

 -->  --> 

On a tablet (iPad Mini sized):

If you decline the installation prompt:

On older mobile browsers where automatic install isn't supported, only the manual directions are shown. (And yes the different capitalization is intentional, that's exactly how it's displayed on iOS vs Android, little cursed fun fact for you lmao)

And once PWA is installed, the option disappears. It also only appears on mobile devices, since there is little point to installing the PWA on Desktop.

# Original PR description below

This builds on the work done by Pecacheu in #1024, but completely reworks the UX.

No modal window is shown. Instead, a new "Install" option has been added to the Settings menu.

On browsers that support the `beforeinstallprompt` event, this page contains a button to install the PWA. Clicking this button starts the PWA install experience. This also gracefully handles the case where the user declines the installation. For browsers which do not support the `beforeinstallprompt` event, manual installation instructions are shown. The install button is hidden within the PWA itself. The install button also only appears on mobile browsers, since currently on desktop I think there is little point in installing the PWA.

~~I also updated the way we use Vite PWA to follow their best practices for SW updates: <https://vite-pwa-org.netlify.app/guide/periodic-sw-updates.html#handling-edge-cases>~~

- `main` 
  - \#1380 :point\_left:

### DIFF
--- a/packages/client/components/app/interface/settings/InstallInstructions.tsx
+++ b/packages/client/components/app/interface/settings/InstallInstructions.tsx
@@ -37,14 +37,13 @@ const Steps = styled("ol", {
  * Installation Instructions Page
  */
 export default function InstallInstructions() {
-  const { pwaPrompt, pwaInstalled } = useDevice();
+  const { pwaPrompt, pwaInstalled, isFirefox, isIOSTouch } = useDevice();
 
   const [result, setResult] = createSignal<"accepted" | "dismissed">();
-  const isFirefox = navigator.userAgent.toLowerCase().includes("firefox");
 
-  createEffect(() => {
-    pwaPrompt()?.userChoice?.then((choice) => setResult(choice.outcome));
-  });
+  createEffect(() =>
+    pwaPrompt()?.userChoice?.then((choice) => setResult(choice.outcome)),
+  );
 
   return (
     <Container>
@@ -101,23 +100,25 @@ export default function InstallInstructions() {
           </h1>
         </Show>
 
-        <h2>
-          <Trans>Android</Trans>
-        </h2>
-        <Steps>
-          <li>
-            <Trans>Open this page in Chrome</Trans>
-          </li>
-          <li>
-            <Trans>Press the</Trans> ⋮ <Trans>button</Trans>
-          </li>
-          <li>
-            <Trans>Tap "Install app" or "Add to Home screen"</Trans>
-          </li>
-          <li>
-            <Trans>Follow the prompts</Trans>
-          </li>
-        </Steps>
+        <Show when={!isIOSTouch}>
+          <h2>
+            <Trans>Android</Trans>
+          </h2>
+          <Steps>
+            <li>
+              <Trans>Open this page in Chrome</Trans>
+            </li>
+            <li>
+              <Trans>Press the</Trans> ⋮ <Trans>button</Trans>
+            </li>
+            <li>
+              <Trans>Tap "Install app" or "Add to Home screen"</Trans>
+            </li>
+            <li>
+              <Trans>Follow the prompts</Trans>
+            </li>
+          </Steps>
+        </Show>
 
         <h2>
           <Trans>iOS / iPadOS</Trans>
--- a/packages/client/components/app/interface/settings/InstallInstructions.tsx
+++ b/packages/client/components/app/interface/settings/InstallInstructions.tsx
@@ -1,6 +1,6 @@
 import { createEffect, createSignal, Match, Show, Switch } from "solid-js";
 
-import { Trans } from "@lingui-solid/solid/macro";
+import { Trans } from "@lingui/solid/macro";
 import { styled } from "styled-system/jsx";
 
 import { useDevice } from "@revolt/common";
--- a/packages/client/components/app/interface/settings/InstallInstructions.tsx
+++ b/packages/client/components/app/interface/settings/InstallInstructions.tsx
@@ -1,0 +1,152 @@
+import { createEffect, createSignal, Match, Show, Switch } from "solid-js";
+
+import { Trans } from "@lingui-solid/solid/macro";
+import { styled } from "styled-system/jsx";
+
+import { useDevice } from "@revolt/common";
+import { Button, Symbol } from "@revolt/ui";
+
+const Container = styled("div", {
+  base: {
+    padding: "var(--gap-lg) 0",
+    color: "var(--md-sys-color-on-surface)",
+    fill: "var(--md-sys-color-on-surface)",
+
+    "& h1": { fontSize: "22px" },
+    "& h2": { fontSize: "18px" },
+
+    "& p, & h1, & h2": {
+      marginBlockStart: "1.2em",
+      marginBlockEnd: ".5em",
+    },
+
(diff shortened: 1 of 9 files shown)


Dadadah:
can you update the screenshots to this pr to feature the latest version (with the mobile branch) to allow ui/ux review

Pecacheu:
@Dadadah Screenshots added!

packages/client/components/app/interface/settings/InstallInstructions.tsx
```diff
@@ -0,0 +1,150 @@
...
+import MdIosShare from "@material-design-icons/svg/outlined/ios_share.svg?component-solid";
+import MdRefresh from "@material-design-icons/svg/outlined/refresh.svg?component-solid";
+import { useDevice } from "@revolt/common";
+import { Button, iconSize } from "@revolt/ui";
+import { createEffect, createSignal, Match, Show, Switch } from "solid-js";
+import { styled } from "styled-system/jsx";
+
+const Container = styled("div", {
+  base: {
+    padding: "15px 0",
```
Dadadah: Please use one of our gap vars here.

packages/client/components/app/interface/settings/InstallInstructions.tsx
```diff
@@ -0,0 +1,150 @@
+import { Trans } from "@lingui-solid/solid/macro";
+
+import MdIosShare from "@material-design-icons/svg/outlined/ios_share.svg?component-solid";
+import MdRefresh from "@material-design-icons/svg/outlined/refresh.svg?component-solid";
```
Dadadah: Please use `<Symbol>` on new pages.

packages/client/fontpreload.plugin.ts
```diff
@@ -29,7 +29,6 @@ export function addFontPreload(): Plugin {
   return {
     name: "vite-add-font-preload",
     transformIndexHtml: {
-      order: "post",
```
Dadadah, line 32: Is this related to this PR? Seems like an unrelated bug fix, perhaps it can be in it's own PR even if it's small so that we get a separate patch note.

Dadadah (changes requested):
I'm not sure we need to include installation instructions on this page - what if we instead put up a support article for installation and just linked to that? I feel that might match the behavior of similar systems better. This may be a UI/UX issue.

Is it really worth importing `vite-plugin-html-minifier`? You're talking a savings of like a few bytes as it won't even minifiy the event names, I don't think we need to import 90 packages for this.

packages/client/fontpreload.plugin.ts
Pecacheu: Oh yeah, that's related but it's a bit hard to tell why. The reason is because when adding the HTML minification plugin (to minify it now that there's some inline JS needed in the HTML file), this other custom plugin that also modifies the HTML file was going out of order with it. Minifiers need to run last or else formatting gets messed up. This made it run in the correct sequence by setting it to default order. Default is what we want here as it also makes it not run too early before the `order: pre` plugins either

packages/client/components/app/interface/settings/InstallInstructions.tsx
Pecacheu: Gotcha. In my defense this is part of Miso's code I think x3

Pecacheu:
> I'm not sure we need to include installation instructions on this page - what if we instead put up a support article for installation and just linked to that? I feel that might match the behavior of similar systems better. This may be a UI/UX issue.

Well someone would need to create that but I think either approach would be fine, as long as it's easy to understand like the writeup in this PR is.

Hmm then again, might be harder for the user to follow it given they'd be doing this on a phone specifically, so avoiding making them go open up an external link on their little phone browser just to read some directions might make more sense vs on PC.

> Is it really worth importing `vite-plugin-html-minifier`? You're talking a savings of like a few bytes as it won't even minifiy the event names, I don't think we need to import 90 packages for this.

That is a good point lol, didn't think about how heavy that dep might be compared to the small savings given our single HTML file. On the plus side it seemed to add no precipitable change to build time.

packages/client/components/app/interface/settings/InstallInstructions.tsx
```diff
@@ -0,0 +1,152 @@
...
+});
+
+/**
+ * Installation Instructions Page
+ */
+export default function InstallInstructions() {
+  const { pwaPrompt, pwaInstalled } = useDevice();
+
+  const [result, setResult] = createSignal<"accepted" | "dismissed">();
+  const isFirefox = navigator.userAgent.toLowerCase().includes("firefox");
```
Dadadah: Should this go into the Device hook? We have an isIosTouch check in there.

packages/client/components/app/interface/settings/InstallInstructions.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <li>
+            <Trans>Tap "Install app" or "Add to Home screen"</Trans>
+          </li>
+          <li>
+            <Trans>Follow the prompts</Trans>
+          </li>
+        </Steps>
+
+        <h2>
+          <Trans>iOS / iPadOS</Trans>
```
Dadadah: Can we use the useragent to detect android vs ios? We have isIosTouch in the Device hook and we can probably rely on that.

Dadadah (changes requested):

Dadadah:
Sorry i didn't catch these earlier but I think these features I'm mentioning are new anyway. Once you address these two I think we can get this to UI review.

Btw were you going to remove the .html file minification?

packages/client/components/app/interface/settings/InstallInstructions.tsx
Pecacheu: Good point. I moved the PWA feature detection into Device in the refactor but that should be in Device too

packages/client/components/app/interface/settings/InstallInstructions.tsx
Pecacheu: Ok. Decided to make it only hide the Android instructions and show both if not because if it's true, we know it's iOS for sure, but if it's false... who knows what it could be. Android (which has like zero navigator string consistency btw), iOS running a weird browser, or some other bizarre OS.

Pecacheu:
> Btw were you going to remove the .html file minification?

Idk, I don't like the idea of having unminified code in production but I guess it's not doing too much harm when it's only one file that's only ~1 KB

Pecacheu:
Ok I at least disabled formatting of <script> tag in HTML so that we could use the minified version of the JS inside the HTML, this saves 200 bytes and more importantly, allows me to sleep at night

Dadadah (approved):

PR #1389, opened 2026-07-25
fix: Make copy file link and open file use preview url

### LINKED ISSUE #1322: Copy File Link adds /original to a file, which prevents embedding when sharing links
### What happened?

When using Copy File Link on a file uploaded to Stoat, the result has /original at the end of the URL. 
This is a link to the file, but in the least usable way. If this link instead does not have /original at the end then files are embedded. Files can then still be downloaded from the embed, so there should be no downside to this. 

For example: I upload an image to a server. I then also want to share this image with other servers I'm in.
With the current behaviour, this image does not embed anywhere I would share this link. Resulting in needing multiple uploads. 
With removing /original from the URL, this image now embeds. Meaning only one upload can be shared in multiple places.

Dadadah:
Previously, copying the file link would cause it to fail to embed when shared again. Changing it to the preview url allows it to embed.

Fixes #1322

No UI Changes.

## How was this PR tested?

Right clicked copied some files and ensured the url was as expected.

## Checklist:

None

- `main` 
  - \#1389 :point\_left:

### DIFF
--- a/packages/client/components/app/menus/MessageContextMenu.tsx
+++ b/packages/client/components/app/menus/MessageContextMenu.tsx
@@ -144,14 +144,14 @@ export function MessageContextMenu(props: {
    * Opens the file preview in a new tab
    */
   function openFile() {
-    window.open(props.file?.originalUrl, "_blank");
+    window.open(props.file?.previewUrl, "_blank");
   }
 
   /**
    * Copies the link to the original url of the file
    */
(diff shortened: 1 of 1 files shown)


Pecacheu (approved):
Well those method names are bass awkwards lmao but I checked the code and indeed the "preview" is the one that doesn't append /original so this fix should work.

PR #1404, opened 2026-07-28
fix: You can click user card buttons again

Dadadah:
Another attempt at fixing the user card buttons

- `main` 
  - \#1404 :point\_left:


Pecacheu (approved):
It worksss

PR #1407, opened 2026-07-28
fix: No longer logs out in dev sometimes

Dadadah:
- `main` 
  - \#1407 :point\_left:


Pecacheu (approved):
Looks good. The changes in Controller.ts are pretty much already implemented in multiuser PR though so it's not the cause of the 'session randomly disappearing' issue but that's a separate problem

PR #1413, opened 2026-07-29
fix: Sometimes I18n provider would load before activate

Dadadah:
I don't know what's causing it, but sometimes the indexdb fails to load and that combined with a race condition in I18N loading caused a permanent empty screen. This PR makes absolute sure that activate is called before the I18N provider renders.

## Checklist:

None

- `main` 
  - \#1413 :point\_left:


Pecacheu:
Do you know why that `initI18n` function is exported? I used VSCode "Check for references" but it didn't find any, and neither did a search.

I would imagine we'd want to actively prevent anyone else besides this context from activating it so exporting that seems like a bad idea.

Dadadah:
Good point. Honestly it doesn't even need to be a function it can just live on the provider component.

Pecacheu (approved):

PR #1442, opened 2026-08-03
fix(instance): Docker build injection fail

Pecacheu:
Should fix the current affliction affecting `beta.stoat.chat`, which is caused not by any issue in the source code, but by the tree shaking and/or minification during build process being too "smart" for its own good and literally optimizing out certain environment variables, combing the object that contains them with the export object of `env.ts` itself, and messing things up when `inject.js` tries to do its thing.

The problem is averted by inserting *any* function code between the import of the meta vars and the exported object in env.ts, causing the build system to decide its too risky to optimize because its behavior is not 100% deterministic at build-time. (It never was deterministic at build-time though, build system just went full Joker on us.)

## How was this PR tested?

By pulling out most of my fur in rage

## Checklist:

Checklists can't save you

Where is your LLM god now

- `main` 
  - \#1442 :point\_left:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,13 +66,4 @@ COPY --from=builder /build/packages/client/dist ./dist
 
 EXPOSE 5000
 
-# Runtime env vars (overridden by Helm chart / docker run)
-ENV VITE_HOST=""
-ENV VITE_API_URL=""
-ENV VITE_DEV_WS_URL=""
-ENV VITE_DEV_MEDIA_URL=""
-ENV VITE_DEV_PROXY_URL=""
-ENV VITE_DEV_GIFBOX_URL=""
-ENV VITE_RNNOISE_WORKLET_CDN_URL=""
-
 CMD ["npm", "start"]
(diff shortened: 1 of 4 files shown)


Dadadah:
Can we start with just removing the env variables from the docker container first? I am not convinced that the compiler is doing anything wrong here.

Pecacheu:
> Can we start with just removing the env variables from the docker container first? I am not convinced that the compiler is doing anything wrong here.

I've already tried that on its own with a full build, it has no effect. I've just left it in because it's a good idea anyways, those vars don't need to be set because undefined should be handled equally to an empty string now.

Also, I am 100% certain that the build chain -somewhere- is at fault because I've literally decompiled the code it generated and found it doesn't match our source code, i.e. it optimized out checks it shouldn't be optimizing out. I'm not aware of a way with Vite/esbuild currently to tell them to ignore optimizing specific constants. I did test excluding the entire env.ts file from optimizeDeps, but this did not help, it seems to optimize the entire thing afterwards regardless.

I think that he only other solution to this problem aside from this is to change how inject.js works completely.

Dadadah (approved):

PR #1459, opened 2026-08-07
fix: don't crash rendering messages that produce no markdown output

### LINKED ISSUE #459: bug: reference links that don't point to anything completely hangs the client
### What happened?

I was reading an essay by John MacFarlane about Markdown and the difficulties that CommonMark project had to go through when implementing certain features when I found out about [reference links].

Obviously, I tried using the same syntax on Revolt for Web to see if they worked, which they do (Screenshot attached below).

Image

But sending a message that only contains a reference link that points to nothing, the client completely hangs and the user becomes unable to do the following actions:
- Reload the client
- Switch channels (Except voice channels)
- Switch servers
- Receive messages

[reference links]: https://spec.commonmark.org/0.31.2/#reference-link

GalaxyAkiii:
Messages consisting only of link reference definitions (e.g. `[a]: b`) parse to a hast tree with no children, since remark-rehype treats definitions as metadata and emits nothing for them.

`injectEmojiSize` accessed `children?.[0].properties` on that tree. That safely handles `children` being missing, but not an empty array, so it threw a TypeError. Because the Markdown component renders without an ErrorBoundary around the message list, the exception crashed the entire list: the message wouldn't show up and the UI locked up.

This is why the issue kept happening even after deleting the message. A failed send stays in the local outbox, gets restored as an unsent draft on reload, and hits the same rendering code again, crashing the app on startup. That's especially annoying on desktop where clearing local storage isn't straightforward. With this fix, the stranded draft renders fine so you can resend or delete it.

It's also pretty bad from a security standpoint, anyone can drop a link reference definition in a public channel and freeze the client for everyone who opens it.

Fixes #459

## How was this PR tested?

- [x] Sent `[a]: b` in a channel on a dev build against prod. Before this fix, the client froze and the message disappeared; now it sends fine and the channel works normally.
- [x] Retried a draft stuck from a previous crash; it sent correctly and cleared from the outbox.
- [x] Verified that empty-rendering content (single/stacked link definitions, zero-width spaces, whitespace, empty messages) renders empty instead of throwing.
- [x] Verified regular text, broken references like `[foo]`, and emoji sizing still render properly.
- [x] Ran `tsc --noEmit`, `eslint`, and `prettier --check` (all passed).

Tested on web. Desktop uses the exact same rendering logic so it should be fine, but I haven't tested it directly on desktop.

N/A - link definitions produce no HTML under CommonMark anyway, so there's no visual difference. It just stops crashing the view.

</details>

## Checklist:

None.

### DIFF
--- a/packages/client/components/markdown/emoji/util.ts
+++ b/packages/client/components/markdown/emoji/util.ts
@@ -36,9 +36,11 @@ export function injectEmojiSize(
   const content = props.content ?? "";
 
   // inject emoji size information
+  // (there is no first child when content renders to nothing,
+  //  such as a message containing only link reference definitions)
   const properties = (
     hastNode as { children?: { properties: Record<string, string> }[] }
-  ).children?.[0].properties;
+  ).children?.[0]?.properties;
 
   // inject custom property
   if (properties) {


Dadadah (approved):

Dadadah:
Thanks for your contribution.

PR #1460, opened 2026-08-07
fix: Fix memory leak w/ new timer system in State

Pecacheu:
I'm sure this is not the only memory leak we have but it should at least improve performance not constantly registering and deregistering so many damn setIntervals.

We also don't need to regenerate relative timestamps literally every second. These only change at time boundaries and it's not a big deal for them to have a delay. Even Discord takes 30+ seconds for your message to go from "Sent Today" to "Sent Yesterday" when the clock strikes midnight.

## How was this PR tested?

Left stoat open for a while in foreground and background to check performance and memory usage.

## Checklist:

You can ask my favorite chemical compound, Sodium Hypobromite. Also known as NaBrO

- `main` 
  - \#1460 :point\_left:

### DIFF
--- a/packages/client/components/state/index.tsx
+++ b/packages/client/components/state/index.tsx
@@ -3,6 +3,7 @@ import {
   Show,
   createContext,
   createSignal,
+  onCleanup,
   onMount,
   useContext,
 } from "solid-js";
@@ -53,6 +54,7 @@ export class State {
   private store: Store;
   private setStore: SetStoreFunction<Store>;
   private writeQueue: Record<string, number>;
+  _timers: (() => void)[] = [];
 
   appDrawer;
   setAppDrawer;
@@ -210,6 +212,17 @@ export class State {
       store.hydrate();
     }
   }
+
+  /** Add an interval timer that runs once per minute, but is cleaned up on logout */
+  perMinute(func: () => void) {
+    this._timers.push(func);
+  }
+
+  /** Remove a timer added with `setPerMinute` */
+  clearPerMinute(func: () => void) {
+    const i = this._timers.indexOf(func);
+    if (i !== -1) this._timers.splice(i, 1);
+  }
 }
 
 /**
@@ -221,13 +234,20 @@ const stateContext = createContext<State>(null! as State);
  * Mount state context
  */
 export function StateContext(props: { children: JSX.Element }) {
-  const stateLocal = new State();
+  const state = new State();
   const [ready, setReady] = createSignal(false);
+  let tmr: NodeJS.Timeout;
 
-  onMount(() => stateLocal.hydrate().then(() => setReady(true)));
+  onMount(() => {
+    state.hydrate().then(() => setReady(true));
+    tmr = setInterval(() => {
+      for (const fn of state._timers) fn();
+    }, 6e4);
+  });
+  onCleanup(() => clearInterval(tmr));
 
   return (
-    <stateContext.Provider value={stateLocal}>
+    <stateContext.Provider value={state}>
       <Show when={ready()}>{props.children}</Show>
     </stateContext.Provider>
   );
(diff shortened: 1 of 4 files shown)


Dadadah (changes requested):
Please migrate this PR to use [@solid-primitives/date](https://primitives.solidjs.community/package/date) instead of a signal we update ourselves. You can host a solid-primitives/date object from the state, or create a global hook like `useDurationFormat` or `useTime`. If you use a global hook, it should probably go in the common folder.

For now, lets keep the times updating once a second. Once a second isn't that often from a computing standpoint, so lets keep it like that until we have evidence it's causing performance issues.

Pecacheu:
I mean, I have evidence it's causing performance issues, namely that switching away from and back to the tab (after at least several mins in background) now doesn't seem to cause nearly as massive of a lag spike as it always has lol. And a timer once a second is a big deal when it's potentially updating hundreds of date objects (every message on the entire currently rendered channel). Especially for something that is only expected to change once or twice _per day._

I switched everything to use the reactive var from solid-primitives/date though, lemme know if that works.

Dadadah:
> I mean, I have evidence it's causing performance issues, namely that switching away from and back to the tab (after at least several mins in background) now doesn't seem to cause nearly as massive of a lag spike as it always has lol. And a timer once a second is a big deal when it's potentially updating hundreds of date objects (every message on the entire currently rendered channel). Especially for something that is only expected to change once or twice _per day._
> 
> I switched everything to use the reactive var from solid-primitives/date though, lemme know if that works.

are you sure the slowdown isn't caused specifically by the intervals? You could try turning it back down to 1s and see if the slowdown is still there with the new reactive date?

Edit: Just do a cpu flame graph with 1s vs 1min and do a page refresh and a tab back and forth and see how it looks. I just wanna make sure it's actually the cause of the slowdown.

packages/client/components/state/stores/NotificationOptions.ts
```diff
@@ -76,13 +73,8 @@ export class NotificationOptions extends AbstractStore<
...
     super(state, "notifications");
 
-    // memory leak? -- maybe this should be a global util somewhere
-    // todo: refactor
-    const [now, setNow] = createSignal<number>(+new Date());
-    this.#now = now;
-
     // update every minute
-    setInterval(() => setNow(+new Date()), 6e3);
+    this.#now = createMemo(() => +state.datePerMinute());
```
Dadadah, line 77: This value is already memoized, I don't think it needs to be memoized again. Maybe just have it set as `this.#now = datePerMinute`, or maybe we can get rid of the #now variable entirely and just set it directly.

Pecacheu: Kinda subtle but the `+` symbol gets the UTC time out of the date object by force-converting it to a number, so there is actually a calculation going on here. That's how they were doing it before but maybe we should use `.getTime()` to be more explicit, that does the same thing but it's less hacky. (Actually, I wonder if this is a trick minifiers use)

packages/client/components/state/index.tsx
```diff
@@ -59,6 +62,18 @@ export class State {
...
+  datePerMinute: Accessor<Date> = createDateNow(6e4)[0];
+
+  /** A reactive Date() that updates only when the day changes */
+  datePerDay: Accessor<Date> = (() => {
+    const poll = createDateNow(1000)[0];
+    const [get, set] = createSignal();
+    createEffect(() => set(poll().getDay()));
+    const date = createMemo(() => (get(), new Date()));
+    return date;
+  })();
```
Dadadah, line 75: This function is a little confusing/unclear. Could you reformat it to be a hook function like useDurationFormat or useTime instead?

Pecacheu: I thought useXYZ is for contexts though, it's inside of useState already. Do you mean renaming it to `useDatePerDay`?

Dadadah (approved):

Issue #828, opened 2026-02-19
2fa not functional on web app


rocielrociel:
### What happened?

2fa is not functional on web app, I can't login, can't enable 2fa, specifically on the web app. 
The desktop client works fine, but the web app returns `invalidtoken` when trying to enable 2fa. And when I set up 2fa elsewhere and try to log in on the web app, it returns no error but keeps booting me back to the 2fa popup.
I use Aegis Authenticator and Librewolf browser, and though I have no idea what the issue might be, the last time I had issue with 2fa on another service it turned out to be a timezone desync. Just leaving that info here, maybe it'll be useful for the team.

GalaxyAkiii:
Some additional context after looking through the MFA flow while working on #810.

**Where the error comes from.** Authenticator setup on the client is a thin pass-through: `POST /auth/mfa/totp` with an MFA ticket to generate the secret, then `PUT /auth/mfa/totp` with `{ totp_code }` to enable it. `InvalidToken` is returned by `enable_totp` in `crates/core/database/src/models/accounts/model.rs` when the submitted code doesn't match the generated one, so nothing in `for-web` is producing it.

**Desktop vs. web.** Worth noting the desktop client has changed since this was filed. At the time it loaded `https://beta.revolt.chat` by default; that was switched to `https://stoat.chat/app` in stoatchat/for-desktop#196 on 6 April. I don't know whether those were equivalent at the time, but it does mean the desktop build compared against here isn't the one anyone would be running now.

**Error visibility.** Not the cause, but until #810 is fixed the setup modal reopens over the error dialog and covers it, so the failure can't be read in the UI. Same PR fixes that.

Given the desktop client now points somewhere different, this is worth re-testing against a current build.


Edit: After verifying against some notes and not being on two hours of sleep when I drafted this I have corrected some mistakes.

PR #1463, opened 2026-08-08
fix: don't reopen the authenticator setup modal over its own error

### LINKED ISSUE #810: "Enable Authenticator" Fails
### What happened?

Every time I try to enable authenticator on my account, be it on the web, or the desktop app, it never wants to work.

It'll look like it works, and error message pops up, and then the authenticator setup screen pops up quickly in rapid succession, so you can't see the error message.

GalaxyAkiii:
`setupAuthenticatorApp` wrapped the TOTP modal in a `while (!success)` retry loop, which led to a couple of broken UX behaviors.

If a code was rejected, `enableAuthenticator` threw, triggering `showError` right before the loop immediately reopened the modal on top of it. The error dialog got covered instantly, so to the user, the setup screen just looked like it flashed and reappeared with zero feedback.

Dismissing the modal had a similar issue. The retry loop only broke on explicit cancellation (which rejected with `MFACancelled`). Closing the dialog via Escape, backdrop click, or the close button routed through `onClose` and called `callback()` with no arguments. Because this resolved to `undefined`, `success` remained falsy and the modal reopened right away, locking the user in the setup flow unless they specifically clicked "Cancel".

`MFAEnableTOTP` actually already contained a `try`/`catch` around `props.callback(code)` meant to display `showError` while keeping the dialog open. However, because `callback` was just `resolve`, it was unreachable and never threw.

This fix passes the enable handler directly into the modal so that catch block runs as intended, allowing us to drop the retry loop entirely. An invalid code now displays the error while keeping the modal open and the input field preserved. Dismissing the dialog via any method now cleanly cancels setup instead of looping.

Fixes #810
Related to #828, which reports InvalidToken when enabling 2FA. This doesn't fix that, the error originates server-side, but it does make the failure readable instead of flashing past, so a rejected code is distinguishable from the modal simply looping.

## How was this PR tested?

- [x] Entered an invalid code: verified the error message stays visible and readable with the entered code preserved in the input field (previously flashed and reset the modal).
- [x] Dismissed via backdrop click: verified it cancels cleanly and shows the "MFA setup canceled" snackbar (previously re-prompted endlessly).
- [x] Dismissed via Escape key: closes cleanly.
- [x] Clicked Cancel button: works as expected and shows the cancellation snackbar.
- [x] Entered a valid code: successfully completes setup.
- [x] Ran `tsc --noEmit`, `eslint`, and `prettier --check` (all passed).

Tested on web against prod with a dev build. Desktop loads the same client, so it should behave identically, though I haven't tested it there directly.

N/A - Control flow fix. The key visual change is that the error state remains visible rather than immediately flashing away.

</details>

## Checklist:

None.

### DIFF
--- a/packages/client/components/modal/types.ts
+++ b/packages/client/components/modal/types.ts
@@ -185,7 +185,7 @@ export type Modals =
       type: "mfa_enable_totp";
       identifier: string;
       secret: string;
-      callback: (code?: string) => void;
+      callback: (code?: string) => Promise<void>;
       reject?: (reason?: string) => void;
     }
   | ({
--- a/packages/client/components/app/interface/settings/user/Account.tsx
+++ b/packages/client/components/app/interface/settings/user/Account.tsx
@@ -163,23 +163,23 @@ function MultiFactorAuth() {
     }
     const secret = await ticket!.generateAuthenticatorSecret();
 
-    let success;
-    while (!success) {
-      try {
-        const code = await mfaEnableTOTP(secret, client().user!.username);
+    try {
+      const enabled = await mfaEnableTOTP(
+        secret,
+        client().user!.username,
+        (code) => mfa.data!.enableAuthenticator(code),
+      );
 
-        if (code) {
-          await mfa.data!.enableAuthenticator(code);
-          success = true;
-        }
-      } catch (err) {
-        // If the modal was closed (ie. promise was rejected) quit the loop
-        if (err === "MFACancelled") {
-          snackbar.show({ message: t`MFA setup canceled.` });
-          return;
-        }
-        showError(err);
+      if (!enabled) {
+        snackbar.show({ message: t`MFA setup canceled.` });
       }
+    } catch (err) {
+      if (err === "MFACancelled") {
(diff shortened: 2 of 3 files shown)


packages/client/components/modal/types.ts
```diff
@@ -185,7 +185,7 @@ export type Modals =
       type: "mfa_enable_totp";
       identifier: string;
       secret: string;
-      callback: (code?: string) => void;
+      callback: (code?: string) => void | Promise<void>;
```
Dadadah: Is this modal called anywhere else? If not, can this value just be `Promise<void>`?

GalaxyAkiii: Good catch, it's only used in one place, `setupAuthenticatorApp` in `settings/user/Account.tsx`, via `mfaEnableTOTP`. Narrowed it to `Promise<void>`; `tsc --noEmit`, eslint and prettier all pass. Amended rather than adding a commit since it's a one-line change, so it's a force-push. Thank you for that.

Dadadah (approved):

PR #1470, opened 2026-08-08
refactor: Rewrite member list logic to not hit stores so much

Dadadah:
Rewrite the memberlist logic.

I have discovered that calling a store multiple times is extremely expensive. This rewrite significantly reduces the calls to a store. I measured a 33% reduction in memberlist load time for the lounge (1000 online users at time of testing). Adding a new user takes 4ms, so we shouldn't see any kind of serious performance hit by unlimiting the number of users visible. If we get a server that has more than 1000 online users, we should consider implementing pagination on the backend side.

The biggest slowdown in this process is in stoat.js hydration. We may want to introduce an interval in the syncserver code to allow loading only 100 users every 25ms or something during initial load in to make the UI not freeze when loading users.

No UI changes

## How was this PR tested?

Many before and after profiles with each optimize step

## Checklist:

None

- `main` 
  - \#1470 :point\_left:

### DIFF
--- a/packages/client/components/ui/components/features/texteditor/codeMirrorAutoCompleteSource.ts
+++ b/packages/client/components/ui/components/features/texteditor/codeMirrorAutoCompleteSource.ts
@@ -69,15 +69,21 @@ export function codeMirrorAutoCompleteSource(
       // avoiding using `instanceof`, presumed slow
       const user = ((entry as { user: User })?.user ?? entry) as User;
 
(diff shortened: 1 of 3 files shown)


hyfetch (approved):
lgtm

Pecacheu:
This is awesome but does this also fix the _members list not being reactive on first app load/updating when members join/leave until you switch channels back and forth_ bug, or is that a separate issue?

PR #1471, opened 2026-08-09
fix: Downgrade VITE_HOST exception to error, use an api check

Dadadah:
Fixes beta

## Checklist:

None

- `main` 
  - \#1471 :point\_left:

### DIFF
--- a/packages/client/components/common/lib/env.ts
+++ b/packages/client/components/common/lib/env.ts
@@ -1,5 +1,5 @@
 export const STOAT_HOST = "stoat.chat";
-export const STOAT_API = "https://api.stoat.chat";
+const STOAT_API = "https://api.stoat.chat";
 
 /** App `stoat.json` endpoint format */
 export interface AppConfig {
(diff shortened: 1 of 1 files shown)


packages/client/components/common/lib/env.ts
```diff
@@ -29,15 +29,24 @@ export const normalizeHost = (host: string) =>
     ? STOAT_HOST
     : host;
 
+export const isStoatOfficialAPI = (api: string) =>
```
Pecacheu: I can't think of any place this would be useful to be exported in the app for now since the "is this Stoat" detection is mainly using the host string rather than API string. And if we need to use this elsewhere in the future, we can always mark it as exported later.

Pecacheu (changes requested):
I'm not a huge fan of the `console.error` instead of throwing, but I can see the argument, so I'll go with it lol. Just one change request. Also good catch on `STOAT_API` export being unused.

Pecacheu (approved):

PR #1474, opened 2026-08-10
chore(ci/cd): Docker image will now build on its native os

Dadadah:
Docker build is taking too long in ci/cd lets try to speed it up by making it run on arm natively and by only making it run x64 when in PRs

- `main` 
  - \#1474 :point\_left:


amycatgirl (approved):
I guess this looks correct????? Idk anything about GitHub actions and this could be entirely wrong

:shipit:

PR #1487, opened 2026-08-12
feat: Universal age gate setting per-device & alternate icon for mature channels

### LINKED ISSUE #1596: make different icons for nsfw channels
### Describe your feature

instead of default # symbol nsfw channels should have something different so you can easily identify them without looking to prevent happy little accidents

Pecacheu:
- Makes the age gate confirmation prompt only appear once per device, prevents annoyance of having to click "agree" over and over again when there are many channels
- As an alternative so you can still identify mature channels before clicking, they now have an alternate icon with a warning symbol
- Refactor ServerSidebar to use symbols instead
- Fix symbol prop overrides not all working (eg. weight setting had literally no effect and I don't think anyone noticed because it's rarely used lol)
- Merges changes from #1251

The icon is an inline generated SVG using the material font, so it should be fairly performant, and highly editable. Eg. we could adjust the size or switch to the 'filled' variant of the warning triangle if UI/UX thinks it works better.

## How was this PR tested?

Browsing through numerous servers and clicking on channels

</details>

## Checklist:

Not using AI is as easy as 1, 2, 5, 44, 17, and I'll keep going like this until I reach 100.

- `main` 
  - \#1487 :point\_left:


Dadadah:
I like this change but I think it should be a setting. For example, on the nsfw page it can have a second checkbox for "Remember my decision for all NSFW channels" or something. I would want it to be toggleable in the settings in case I joined a server that had a rather gnarly NSFW channel that I didn't want to see ever, and in that case it'd be useful to be able to turn the "remember" setting off in the settings.

Pecacheu:
> I like this change but I think it should be a setting. For example, on the nsfw page it can have a second checkbox for "Remember my decision for all NSFW channels" or something. I would want it to be toggleable in the settings in case I joined a server that had a rather gnarly NSFW channel that I didn't want to see ever, and in that case it'd be useful to be able to turn the "remember" setting off in the settings.

I think that would make sense, though in that scenario you'd also need a "Hide NSFW channel content" button (or whatever you want to call it) somewhere so that you can "undo" your decision to enter it. Because currently (both in main and in this PR) there's no way to do so- it's a one-way choice, once you agree to view an NSFW channel, that one is unhid forever on that device. (Apart from wiping your store data that is ofc)

Question would just be where to put that button. It would be awkward to have it taking up space on the top navbar at all times when it would be a rarely used option, but you don't want it unfindable either. Perhaps a settings tab with a list of channels you've previously shown, or shove it in the channel right-click context menu?

PR #1506, opened 2026-08-17
feat: join voice automatically when opening a voice channel

valterh4ck3r:
Opening a voice channel already expresses the intent to join it, but today connecting takes a second click on the "Join the voice channel" card. This connects on navigation instead.

`TextChannel` gains an effect keyed on the channel id that calls the same `voice.connect()` the call card's `onClick` already used, so there is no new connection path — only an earlier trigger for the existing one.

Two cases are deliberately excluded:

- **DMs and groups.** `Channel#isVoice` is also true for `DirectMessage` and `Group`, so without the `type === "TextChannel"` guard, opening a conversation would start ringing the other side.
- **Opening a channel while already in a call.** `connect()` disconnects first, so auto-connecting here would drop an ongoing call just because another voice channel was opened to read its chat. The card keeps showing "Switch to this voice channel" and switching stays a manual action.

## How was this PR tested?

- [x] `tsc --noEmit` on `packages/client` (after building `stoat.js` and `solid-livekit-components`) reports no errors from this change; the only remaining error is the pre-existing missing generated lingui catalog.
- [x] `prettier --check` and `eslint` are clean on the changed file.
- [ ] Not exercised at runtime against a live instance — the behaviour itself is unverified and worth a reviewer's attention.

## Checklist:

The patch and this description were drafted by Claude Code (Claude Opus 5), working from my report of the behaviour I wanted. I reviewed the diff and the two scoping decisions above — excluding DMs/groups, and refusing to drop an ongoing call — before opening the PR.

I am aware the contributing guide asks contributors to avoid PRs generated entirely by language models; I am declaring this openly rather than passing it off as hand-written, and I am happy to rework or close it if that is not acceptable here.

### DIFF
--- a/packages/client/src/interface/channels/text/TextChannel.tsx
+++ b/packages/client/src/interface/channels/text/TextChannel.tsx
@@ -16,6 +16,7 @@ import { DraftMessages, Messages } from "@revolt/app";
 import { useClient } from "@revolt/client";
 import { Keybind, KeybindAction, createKeybind } from "@revolt/keybinds";
 import { useNavigate, useSmartParams } from "@revolt/routing";
+import { useVoice } from "@revolt/rtc";
 import { useState } from "@revolt/state";
 import { LAYOUT_SECTIONS } from "@revolt/state/stores/Layout";
 import {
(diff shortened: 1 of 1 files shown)


Dadadah:
We do not accept PRs made with LLMs. Furthermore, this is intentional and by design.

PR #1509, opened 2026-08-18
fix: limit users in voice call preview to 3, like on android

amycatgirl:
The voice call preview card was overflowing when multiple people joined at once. This PR makes it so that
it doesn't overflow when there are too many people inside of it, matching the behaviour present on Stoat for Android.

## How was this PR tested?

- [x] Call card shows "With <user 1>, <user 2>, <user 3>" when there are at least 3 people in the call
- [x] Call card shows "With <no. of users in call>" when there are more than 3 people in the call
- [x] Call card shows "Start the call" when there are no people in the call

## With more than 3 users
image

## With three users or less
image

## With no users in the call
image

</details>

## Checklist:

No AI was used when authoring this pull request.

### DIFF
--- a/packages/client/components/ui/components/features/voice/callCard/VoiceCallCardPreview.tsx
+++ b/packages/client/components/ui/components/features/voice/callCard/VoiceCallCardPreview.tsx
@@ -83,7 +83,7 @@ const Avatars = styled("div", {
     height: "fit-content",
 
     "& :not(:first-child)": {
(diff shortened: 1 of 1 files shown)


Dadadah (approved):
I'm not a fan of this duplicate code but I can fix that later. I think we should create a "stacked avatar" ui element that handles this for us and scales the margin by the size of the avatars.

PR #1529, opened 2026-08-19
fix: unknown users when full member list is not loaded

### LINKED ISSUE #1441: Users joining voice in server that don't load the full userlist show up as UnknownUser in voice
### What happened?

In large servers that do not load the full userlist (lounge) users' data will not be loaded when joining voice. Therefor new voice users will show up as `UnknownUser`

Should be fixable by loading the user when we receive a new participant.

amycatgirl:
Fixes #1441

Should fix users not being completely loaded on the channel sidebar and inside a voice chat (if the user hasn't
sent a message inside the channel yet.)

The only api facing change is that useUsers is now using a resource instead of being a memo, which has the advantage
of allowing async functions while only recalculating the values when the supplied IDs change. It should also handle
the edge case of having a spotty connection and not being able to fetch a user (by falling back to "unknown user".)

Nothing broke on my very limited testing of this change.

## How was this PR tested?

- [x] User loaded when they weren't in cache already 
- [x] Checked that users could load from cache and from the network where we use `useUser` and `useUsers`

</details>

## Checklist:

No AI was used to author this pull request

### DIFF
--- a/packages/client/components/markdown/users.ts
+++ b/packages/client/components/markdown/users.ts
@@ -1,4 +1,8 @@
-import { Accessor, createMemo } from "solid-js";
+import {
+  type Accessor,
+  type InitializedResource,
+  createResource,
+} from "solid-js";
 
 import { ServerMember, User } from "stoat.js";
 
@@ -53,6 +57,39 @@ export function userInformation(user?: User, member?: ServerMember) {
   };
 }
 
+async function getOrFetchUsers(options: {
+  ids: string[];
+  params: Accessor<{ serverId?: string }>;
+  filterNull?: boolean;
+}) {
+  const client = useClient();
+  const serverId = options.params().serverId;
+  return Promise.all(
+    options.ids.map(async (id) => {
+      const user =
+        client().users.get(id) ??
+        (await client()
+          .users.fetch(id)
+          .catch(() => null));
+
+      // Edge case: what if user fails to fetch on a spotty/intermittent conenction?
+      // Promise.all fails as soon as possible, so if one fetch fails, everything else
+      // does too and we don't get any data at all
+      if (!user) return undefined;
+
+      return userInformation(
+        user,
+        serverId
+          ? client().serverMembers.getByKey({
+              server: serverId,
+              user: user.id,
+            })
+          : undefined,
+      );
+    }),
+  ).then((arr) => (options.filterNull ? arr.filter((x) => x) : arr));
+}
+
 /**
  * Resolve multiple users by their ID within the current context
  * @param ids User IDs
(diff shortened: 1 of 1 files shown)


packages/client/components/markdown/users.ts
```diff
@@ -53,6 +57,38 @@ export function userInformation(user?: User, member?: ServerMember) {
...
+          .catch(() => null));
+
+      // Edge case: what if user fails to fetch on a spotty/intermittent conenction?
+      // Promise.all fails as soon as possible, so if one fetch fails, everything else
+      // does too and we don't get any data at all
+      if (!user) return undefined;
+
+      return userInformation(
+        user,
+        options.params().serverId
```
Dadadah: Can you refactor this to only call `options.params()` once, ie. outside of the `Promise.all()` call? Calling an accessor backed by solid in a loop is very expensive.

packages/client/components/markdown/users.ts
```diff
@@ -53,6 +57,38 @@ export function userInformation(user?: User, member?: ServerMember) {
...
+  filterNull?: boolean;
+}) {
+  const client = useClient();
+  return Promise.all(
+    options.ids.map(async (id) => {
+      const user =
+        client().users.get(id) ??
+        (await client()
+          .users.fetch(id)
+          .catch(() => null));
```
Dadadah, line 73: Don't we have a getOrFetch method? Or will that not work here?

packages/client/components/markdown/users.ts
```diff
@@ -53,6 +57,38 @@ export function userInformation(user?: User, member?: ServerMember) {
...
+  ids: string[];
+  params: Accessor<{ serverId?: string }>;
+  filterNull?: boolean;
+}) {
+  const client = useClient();
+  return Promise.all(
+    options.ids.map(async (id) => {
+      const user =
+        client().users.get(id) ??
+        (await client()
```
Dadadah, line 71: Are you sure that this won't throw if it fails to fetch (and return an error, making `Promise.all()` fail)?

Dadadah (changes requested):

packages/client/components/markdown/users.ts
amycatgirl: Apparently not lol

amycatgirl: I think it makes sense to put this into stoat.js proper though

packages/client/components/markdown/users.ts
amycatgirl: Get returns null if there aren't any matches in cache
And the catch catch inside the await returns null so I don't think it would throw?

packages/client/components/markdown/users.ts
amycatgirl: Done in [5496dee](https://github.com/stoatchat/for-web/commit/5496dee616d8df6436864d6c0fc325459097fae9)

packages/client/components/markdown/users.ts
Dadadah: ah I think when I reviewed this I didn't see the catch down below.

Dadadah (approved):

PR #1497, opened 2026-08-15
fix: Always ensure the stream starts as below 720p

Dadadah:
Many users are having an issue where they get kicked from the call when screen sharing. This PR might fix that, though I am unable to replicate it so I don't know for sure.

This PR forces the stream to always start at max 420p no matter what. The stream is then enlarged to the correct size afterwards.

No UI changes

## How was this PR tested?

Started streams and checked they met the expected resolution and framerate

## Checklist:

- [x] I have carefully read [the contributing guidelines](https://developers.stoat.chat/developing/contrib/)
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation if applicable
- [x] I have no unrelated changes in the PR
- [x] I have confirmed that any new dependencies are strictly necessary
- [ ] I have written tests for new code (if applicable)
- [x] I have followed naming conventions/patterns in the surrounding code

## Please declare, if any, LLM usage involved in creating this PR

None

- `main` 
  - \#1497 :point\_left:

Issue #1318, opened 2026-06-30
Feat: Allow customizable Video Quality (Resolution, FPS, and Bitrate) via configuration


wellcoming:
### **Description:**

Currently, video and screenshare streams are heavily limited by hardcoded presets (e.g., maximum 1080p 30FPS and low default bitrates) within the client side (`packages/client/components/rtc/state.tsx`).
### **Suggested Solution:**

Introduce a way to configure more flexible stream qualities.

PR #1530, opened 2026-08-19
fix: publish screen shares at the selected frame rate, add a 1080p60 quality

devoliverluccas:
### The bug

Every screen share goes out at 15fps, whichever quality is selected.

`toggleScreenshare` calls `setScreenShareEnabled` with capture options but no publish options:

```ts
await room.localParticipant.setScreenShareEnabled(true, { resolution, audio });
```

so livekit falls back to `publishDefaults`, and livekit-client sets:

```js
publishDefaults = {
  ...
  screenShareEncoding: ScreenSharePresets.h1080fps15.encoding,  // 2.5Mbps, 15fps
};
```

That is a hard cap on the encoder. The `applyConstraints` call afterwards only touches the capture side, which is why the source runs at 30fps while what arrives at the far end is 15fps — and why the picker can keep saying "1080p 30FPS" while it is not true.

### The fix

Each quality now carries the `VideoEncoding` it wants, and it is passed through on publish.

The encoding is fixed on the sender when the track goes out, so picking a quality afterwards (from the desktop source picker, or the "always ask" modal) also has to push it onto the sender directly via `setParameters` — otherwise only the value already in settings would ever apply. Simulcast layers are scaled-down copies of the same picture, so the bitrate is divided by the square of `scaleResolutionDownBy` and only the full-resolution layer gets the whole budget.

Four smaller things in the same path:

- **`degradationPreference` is now set, per quality.** Nothing was requested before, so frame rate was the first thing shed under pressure. Motion qualities now ask for `maintain-framerate` and the text quality for `maintain-resolution`, matching the `contentHint` each already sets — the text quality exists to keep small text readable, so it should give up frames, not sharpness.
- **The height constraint was built from the width.** For 1080p it asked for `{ ideal: 1920, max: 1080 }` — an ideal above its own max — and the `=== 0` guard beside it tested `width` as well.
- **`applyConstraints` only set `frameRate.max`.** Capture starts clamped to `{ ideal: 5, max: 5 }` by the `getDisplayMedia` override added in #1497, and a bare `max` does not ask the source to speed back up, so `ideal` is sent too.
- **`getEnabledScreenShareQualities` wrote through to `ScreenSharePresets.original.resolution`.** That object is a singleton exported by livekit-client, so the preset stayed corrupted for the rest of the session. It builds its own object now.

### The new quality

A "1080p 60FPS" option, for sharing where motion matters more than detail. livekit ships no 60fps screen share preset, so the encoding is built from the 1080p30 one at 6Mbps — a higher frame rate needs more budget to stay legible while moving, or the encoder trades the frame rate straight back away and there is nothing to gain over 1080p30. It sits behind the same server-side resolution check that gates 1080p30.

Related to #1318, which asks for configurable quality but does not mention that the current options do not deliver what they advertise.

### Heads up on load

This does increase real bandwidth use. Today every share is capped at 2.5Mbps/15fps; afterwards "1080p 30FPS" actually uses the 5Mbps it promises. That is the intended behaviour, but it is a genuine change in server load — so if you would rather take the correctness fix on its own, I am happy to split the 60fps quality out into a separate PR.

### Testing

`mise build:check`, `mise lint` and `mise format` all pass.

I have **not** verified this at runtime — no measured fps in a real call — so confirmation from someone able to check the outbound `framesPerSecond` in `getStats()` before and after would be worth having.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/client/components/rtc/state.tsx
+++ b/packages/client/components/rtc/state.tsx
@@ -70,6 +70,25 @@ type ScreenShareQuality = {
   contentHint: string;
 };
 
+/**
+ * Frame rate for the "smooth" screen share quality.
+ *
+ * livekit-client ships no 60fps screen share preset, so this one is built by
+ * hand from the 1080p30 preset.
+ */
+const SMOOTH_FRAME_RATE = 60;
+
+/**
+ * Bitrate for the "smooth" screen share quality.
+ *
+ * A higher frame rate needs more bitrate to stay legible while the picture is
+ * moving, or the encoder trades the frame rate straight back away and there is
+ * nothing to gain over 1080p30. 60fps does compress better per frame than
+ * 30fps, so this sits above the 5Mbps of 1080p30 without doubling it, and
+ * stays under the 7Mbps the `original` preset already asks for.
+ */
+const SMOOTH_MAX_BITRATE = 6_000_000;
+
 class Voice {
   #settings: VoiceSettings;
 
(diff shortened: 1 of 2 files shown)


Dadadah:
We do not accept contributions written by LLMs. Please read our contribution guidelines.

PR #1534, opened 2026-08-20
fix: stop DM unread badges from returning on restart

viniciuslrangel:
## Summary
- Bump `stoat.js` to persist optimistic channel acks in `localStorage`, merge them on `/sync/unreads`, and flush debounced acks on page hide
- Ack open channels when the page is visible (not only `document.hasFocus()`), so Electron custom frames still persist reads

Root cause: mentions write to Mongo synchronously, but acks go Redis → AMQP → crond → Mongo. When that pipeline drops/lags, a DM can clear in-session (optimistic UI) and come back unread on every cold start. Observed on VPS with stuck Redis `acker:*` keys and a DM unread doc that had mentions but no `last_id`.

## Test plan
- [ ] Open the sticky DM, confirm badge clears
- [ ] Fully quit desktop app, reopen — DM stays read
- [ ] Receive a new DM, confirm unread still appears; open it; restart — stays read


viniciuslrangel:
Opened against wrong remote; fix lives on viniciuslrangel/stoat-for-web fork.

PR #1536, opened 2026-08-20
fix(voice): opt-in screen share watching on join

viniciuslrangel:
## Summary
- Remote screen shares (video + audio) no longer auto-subscribe/play when joining a voice channel or when a new share appears.
- Reuses existing Stop watching / Resume watching UX: tiles start in the paused/play-to-watch state until the user opts in.
- Mic/camera subscription behavior is unchanged.

## Test plan
- [ ] Join a voice channel where someone is already screen sharing → share tile shows play overlay, no video/audio until Resume watching / tile click
- [ ] While in call, another user starts sharing → same paused state; Resume watching starts video + share audio
- [ ] Stop watching still unsubscribes; Resume watching works again
- [ ] Remote mic/camera tiles still appear and auto-play as before
- [ ] Local screen share preview still works (not stuck in paused overlay)


viniciuslrangel:
Opened against upstream by mistake; fix belongs on viniciuslrangel/stoat-for-web.

PR #1537, opened 2026-08-20
fix(voice): stop multi-viewer screen share flicker

viniciuslrangel:
## Problem
Watched screen shares flickered when several people were in the call. `VideoTrack` with `manageSubscription` unsubscribed any tile under 80% visibility (3s debounce). Crowded grids shrink tiles, so watched shares churned subscribe/unsubscribe.

## Fix
- Never use visibility-based `manageSubscription` for screen share (main tile + PiP). Opt-in watch (`resumeWatching` / `stopWatchingLive`) owns subscription only.
- Skip redundant `setSubscribed` when already at the desired state (audio manager + pause/resume helpers).
- Stop bumping the watching epoch on every pause call (video then audio `TrackPublished`).

Publisher simulcast was already empty; LiveKit `adaptiveStream`/`dynacast` default off. No publisher change needed.

## Verify
1. Share a screen (prefer gaming/high).
2. Two+ clients join and Resume watching.
3. Keep the share focused or in a multi-tile grid; video should stay steady (no blank/flash loop).
4. Stop watching / Resume watching still works; cameras still use manageSubscription.


viniciuslrangel:
Opened against upstream by mistake; belonging on viniciuslrangel/stoat-for-web.

PR #1538, opened 2026-08-20
fix: Normalize regex format

Pecacheu:
Sister PR to <https://github.com/stoatchat/javascript-client-sdk/pull/188>

I left it as A-Z in `routing/index.tsx` because that is validated by one of the other ones first before it reaches that point (this method is for parsing, not validation of link format), and there's just a lot of them so it would look ugly lol.

## How was this PR tested?

Making sure the app starts and channels/users/mentions in messages didn't look broken

Le-*no*-vo

- `main` 
  - \#1538 :point\_left:


Dadadah (approved):

Dadadah:
if you break the beta build with this I'm never letting you hear the end of it

PR #1539, opened 2026-08-20
fix(voice): show channel participants before joining

viniciuslrangel:
## Summary
- Stop collapsing the pre-join call card to height 0 when the mount has not laid out yet (regression from the call/chat split).
- Fill the reserved preview strip and list participants from API voice state in the sidebar and join card before LiveKit connect.

## Test plan
- [ ] Open a voice channel while others are already in the call (do not join yet).
- [ ] Confirm the top join card shows their avatars and names/count.
- [ ] Confirm the server sidebar under that channel lists the same people.
- [ ] Join the call; active room/grid still works. Leave; preview returns with members.
- [ ] Empty voice channel still shows “Start the call”.


viniciuslrangel:
Opened against upstream by mistake; fix lives on viniciuslrangel/stoat-for-web.

PR #1540, opened 2026-08-20
fix(voice): treat screen share volume 0 as silence

viniciuslrangel:
## Summary
- `getScreenShareVolume` / `getUserVolume` used `|| 1.0`, so volume `0` was treated as unset and `RoomAudioManager` applied 100% to LiveKit `AudioTrack` while the MDUI slider stayed at 0.
- Default to 1.0 only when the stored value is missing/non-finite. Dragging to 0 stays silent and does not flip the mute checkbox.

## Test plan
- [ ] Join voice, Resume watching a remote screen share, unmute screen share audio
- [ ] Drag Screen Share Volume to 0: audio silent, slider stays at 0%, mute checkbox unchanged
- [ ] Drag up from 0: audio returns
- [ ] Same check on participant mic Volume slider
- [ ] Mute Screen Share still disables the publication independently of volume


viniciuslrangel:
Opened against upstream by mistake; this change belongs on viniciuslrangel/stoat-for-web.

PR #1549, opened 2026-08-21
fix(voice): drop ghost preview members after leave

viniciuslrangel:
## Summary
- Root cause of sticky pre-join voice members was often **server-side**: LiveKit `participant_left` webhooks to `voice-ingress` return 500 while Mongo segfault-loops, so Redis voice state (and thus API `voice_states`) never drops leavers.
- Client hardening:
  - stoat.js: clear all `voiceParticipants` on Ready before applying `voice_states`; implement `VoiceChannelMove`
  - On End Call / disconnect, delete the local user from `channel.voiceParticipants` immediately so the join card and sidebar preview update even if leave webhooks fail

## Test plan
- [ ] Two accounts: A joins a voice channel; B sees A in sidebar + join card
- [ ] A hits End Call; B’s preview removes A without refresh
- [ ] A rejoins then leaves quickly while connecting; no ghost self in preview
- [ ] Hard refresh with empty channel shows “Start the call”


viniciuslrangel:
Opened against upstream by mistake

DeclanChidlow:
Do not do this again or we will have no choice but to block you from this GitHub organisation.

viniciuslrangel:
Sorry very much. I mixed the upstream repo in the CLI

PR #1554, opened 2026-08-25
fix: actually fetch the user in the background without blocking ui

amycatgirl:
fixes the fix in #1529 because it was done wrong and introduced a very bad regression

it made ViewBot hang and typing indicators not appear and more stuff

in hindsight, i should've done this from the start
one could say this is a 56 /shrug

## How was this PR tested?

- [x] Typing indicators appear again
- [x] Channels now load without the loading indicator briefly appearing
- [x] Viewing bots no longer completely hangs the client
- [x] The fix fixes the issue that it was supposed to fix (unknown users in voice calls)

</details>

## Checklist:

- [ ] I have performed a self-review of my own code

no ai was used to author this pull request

### DIFF
--- a/packages/client/components/markdown/users.ts
+++ b/packages/client/components/markdown/users.ts
@@ -1,8 +1,4 @@
-import {
-  type Accessor,
-  type InitializedResource,
-  createResource,
-} from "solid-js";
+import { type Accessor, createMemo } from "solid-js";
 
 import { ServerMember, User } from "stoat.js";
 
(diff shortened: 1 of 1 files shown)


packages/client/components/markdown/users.ts
```diff
@@ -99,18 +62,47 @@ async function getOrFetchUsers(options: {
...
-      ids: typeof ids === "function" ? ids() : ids,
-      params,
-      filterNull,
-    },
-    getOrFetchUsers,
-    { initialValue: [] as (UserInformation | undefined)[] },
-  );
+  const clientAccessor = useClient();
+
+  const pending = new Map<string, Promise<unknown>>();
```
amycatgirl, line 70: Not sure if using a Map is the right approach here.

packages/client/components/markdown/users.ts
```diff
@@ -99,18 +62,47 @@ async function getOrFetchUsers(options: {
...
+  function ensureUser(id: string) {
+    if (clientAccessor().users.has(id) || pending.has(id)) return;
+    const promise = clientAccessor()
+      .users.fetch(id)
+      .catch(() => undefined);
+
+    pending.set(id, promise);
+
+    void promise.finally(() => {
```
Dadadah: Why is the void keyword here?

packages/client/components/markdown/users.ts
Dadadah: I think the map here is fine.

Dadadah (approved):

PR #1562, opened 2026-08-28
fix: keep Save file in current tab when supported

StillM8:
## Summary

- keep the current tab when the browser supports the `download` attribute
- preserve the existing `_blank` fallback for browsers that do not support it
- apply the behavior consistently to the context menu, file information, and image viewer download actions
- centralize the browser capability check in `supportsAnchorDownload`

I SOLVED IT, PASSED IT, then AI generated the PR.

This change is intentionally limited to file download behavior.


Dadadah:
Please read our contribution guidelines. We do not accept contributions written with LLMs.

PR #1578, opened 2026-08-31
fix: prevent message draft loss when re-entering a channel through an unwritable one

### LINKED ISSUE #1191: Message drafts may not be restored when navigating between channels
### What happened?

Had to shorten the title a bit.

Basically, Stoat web client has draft messages per channel. Neat.

However, if you navigate away to a channel you don't have send message permissions for, and then navigate back to the original channel where the draft was written, the draft message will not restore (except for currently replying to thing).

It works fine if you navigate from original→writable→original channel. If you navigate from original→unwritable→original, draft isn't restored.

The lost draft can be restored by navigating to a different writable channel, but I had to figure that out through trial and error, so it really looks like losing drafts out of nowhere.

aashish00021:
`MessageComposition` (`Composition.tsx`) stays mounted for the whole session, but `MessageBox.tsx` renders the actual editor (`TextEditor2`) inside a `<Switch>`/`<Match when={!props.sendingAllowed}>` so the editor is genuinely unmounted (not hidden) while viewing a channel you can't send in, and a fresh instance mounts when you return to a writable one.

`TextEditor2` seeds its content once, at construction, from `props.initialValue`. That value was kept in sync via a plain `createEffect(on(() => props.channel, ...))`, which in Solid runs *after* the current render pass, not during it. `sendingAllowed` (which drives the `<Switch>`) is read directly in JSX, so it updates in the same synchronous pass as the channel change. That means when you navigate back to a writable channel, the fresh `TextEditor2` can mount *before* `initialValue` has been updated for the new channel so it seeds itself from whatever was left over from the unwritable channel you just came from (empty), even though the actual draft is still sitting untouched in the draft store.

Fix: switch that one effect to `createRenderEffect`, which runs in the same synchronous render pass as the JSX, closing the gap. The other effect in this file (clearing the draft after a message sends) is untouched it isn't tied to this mount/unmount timing at all.

Fixes #1191 

## How was this PR tested?

- [x] Reproduced the original bug locally (draft → navigate to a channel without send permission → navigate back → draft gone), then confirmed this change fixes it: draft now correctly persists through that navigation.
- [x] `pnpm --filter client exec tsc --noEmit` (same command CI's Typecheck job runs) zero errors in the changed file; the 5 pre-existing errors elsewhere in the repo are unchanged and unrelated to this change.

</details>

## Checklist:

...

### DIFF
--- a/packages/client/src/interface/channels/text/Composition.tsx
+++ b/packages/client/src/interface/channels/text/Composition.tsx
@@ -3,6 +3,7 @@ import {
   Show,
   createEffect,
   createMemo,
+  createRenderEffect,
   createSignal,
   on,
   onCleanup,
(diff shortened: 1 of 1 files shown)


Dadadah:
We do not accept contributions made with LLMs. Please see our contribution guidelines.

Issue #933, opened 2026-02-25
Add folders for servers


Klip29mat2:
### Describe your feature

why d'ont add at discord like you can make folders and put colors in ?

Pecacheu:
Duplicate of #721

PR #1589, opened 2026-09-03
feat: server folders in the server list

### LINKED ISSUE #721: [Feature] Server folders
Would be dope if users could have Folders to group multiple servers in the leftbar

It's especially useful if you have a lot of different servers and it starts to get irritating to find the ones you need

GalaxyAkiii:
Adds collapsible folder grouping to the server list sidebar.

Resolves #721
Addresses stoatchat/stoatchat#577 (and duplicate #933)

### What changes

- **Store & sync (`Ordering.ts`):**
  - Expanded the ordering data model from `string[]` to `(string | ServerFolder)[]`.
  - Stays inside `SynchronisedStores`, so folders sync across devices with zero backend/schema changes.
  - `clean()` handles deduping and migration when dealing with flat arrays written by older clients.
  - `orderedServers()` still returns a flattened array with folder items inlined so keyboard shortcuts, navigation, and the emoji picker don't break.
  - Sidebar consumes the new `orderedEntries()` accessor for the nested layout.
- **Sidebar UI (`ServerList.tsx`):**
  - Split out `ServerEntry` from `ServerList`.
  - Added `FolderEntry` with expand/collapse toggle and aggregated unread/mention counts when folded.
  - Added `FolderGroup` with a theme-aware `color-mix(...)` background so the expanded rail stands out cleanly across themes.
- **Context menus & modals:**
  - Added "Add To folder" submenu and "Remove From folder" to `ServerContextMenu`.
  - Added `ServerFolderContextMenu` to rename or unpack/delete folders.
  - Added `EditServerFolder` modal for renaming.

### Notes & caveats

- **No nested DnD yet:** The current `solid-dnd-directive` setup runs on a single flat container. Teaching it nested drop zones is a much bigger refactor, so adding/removing from folders is context-menu only for now.
- **Older clients:** If an older client without folder support writes to the sync store, its `clean()` step strips non-string IDs. Server order survives, but folders get unpacked back to flat entries on their write.
- **Folder visuals:** Folder colors are in the model and render if set, but there's no color picker in the modal yet. Icon is a plain glyph rather than a grid preview.

## How was this PR tested?

- [x] Created folders, moved servers in/out via context menu
- [x] Deleting a folder unpacks member servers back to the root list in place
- [x] Unread and mention badges bubble up to the folder icon when collapsed
- [x] Tested `color-mix` surface contrast across dark, light, and high-contrast themes
- [x] Keyboard navigation still traverses through all servers in order
- [x] `pnpm exec tsc --noEmit` clean across all feature files
- [x] `pnpm exec eslint` and `pnpm exec prettier --check` clean (0 errors, 0 warnings across all 10 files)

Create Folder/Add Server to existing folder
Screenshot 2026-09-03 at 11 57 26 AM
Screenshot 2026-09-03 at 9 34 43 PM
Folder in server list open
Screenshot 2026-09-03 at 7 35 52 PM
Rename Folder
Screenshot 2026-09-03 at 7 35 25 PM
Folder Management
Screenshot 2026-09-03 at 7 35 09 PM
Folder Closed
Screenshot 2026-09-03 at 9 27 46 PM

</details>

## Checklist:

None

### DIFF
--- a/packages/client/components/app/menus/ServerContextMenu.tsx
+++ b/packages/client/components/app/menus/ServerContextMenu.tsx
@@ -1,3 +1,4 @@
+import MdFolder from "@material-design-icons/svg/outlined/folder.svg?component-solid";
 import { For, Show } from "solid-js";
 
 import { Trans } from "@lingui/solid/macro";
@@ -11,7 +12,9 @@ import { Column, Text, Time } from "@revolt/ui";
 
 import MdAlternateEmail from "@material-design-icons/svg/outlined/alternate_email.svg?component-solid";
 import MdBadge from "@material-design-icons/svg/outlined/badge.svg?component-solid";
+import MdCreateNewFolder from "@material-design-icons/svg/outlined/create_new_folder.svg?component-solid";
 import MdFace from "@material-design-icons/svg/outlined/face.svg?component-solid";
+import MdFolderOff from "@material-design-icons/svg/outlined/folder_off.svg?component-solid";
 import MdLogout from "@material-design-icons/svg/outlined/logout.svg?component-solid";
 import MdMarkChatRead from "@material-design-icons/svg/outlined/mark_chat_read.svg?component-solid";
 import MdNotificationsActive from "@material-design-icons/svg/outlined/notifications_active.svg?component-solid";
@@ -232,6 +235,47 @@ export function ServerContextMenu(props: { server: Server }) {
           </For>
         </ContextMenuSubMenu>
       </Show>
+
+      <ContextMenuSubMenu
+        onClick={() =>
+          state.ordering.createFolder("New Folder", [props.server.id])
+        }
+        icon={MdFolder}
+        buttonContent={<Trans>Add to folder</Trans>}
+      >
+        <For each={state.ordering.folders()}>
+          {(folder) => (
+            <ContextMenuButton
+              onClick={() =>
+                state.ordering.addToFolder(folder.id, props.server.id)
+              }
+            >
+              <Show when={folder.name} fallback={<Trans>Unnamed folder</Trans>}>
+                {folder.name}
+              </Show>
+            </ContextMenuButton>
+          )}
+        </For>
+        <ContextMenuButton
+          icon={MdCreateNewFolder}
+          onClick={() =>
+            state.ordering.createFolder("New Folder", [props.server.id])
+          }
+        >
+          <Trans>New folder</Trans>
+        </ContextMenuButton>
+      </ContextMenuSubMenu>
+
+      <Show when={state.ordering.folderOf(props.server.id)}>
+        <ContextMenuButton
+          icon={MdFolderOff}
+          onClick={() => state.ordering.removeFromFolder(props.server.id)}
+        >
+          <Trans>Remove from folder</Trans>
+        </ContextMenuButton>
+      </Show>
+
+      <ContextMenuDivider />
 
       <ContextMenuSubMenu
         symbol={MdNotificationSettings}
--- a/packages/client/components/state/stores/Ordering.ts
+++ b/packages/client/components/state/stores/Ordering.ts
@@ -1,12 +1,24 @@
-import { Client } from "stoat.js";
+import { Client, Server } from "stoat.js";
 
 import { State } from "..";
 
 import { AbstractStore } from ".";
+import { ServerFolder } from "./ServerFolders";
+
+/**
+ * A resolved entry, ready to render
+ */
+export type ResolvedEntry =
+  | { type: "server"; id: string; server: Server }
+  | { type: "folder"; id: string; folder: ServerFolder; servers: Server[] };
 
 export interface TypeOrdering {
   /**
    * Ordered list of server IDs
+   *
+   * Servers inside folders stay in this list, and a folder is drawn where the
+   * first of its servers sits. Kept as plain ids so that clients without
+   * folder support can read and write this key without losing anything.
    */
   servers: string[];
 }
@@ -22,6 +34,12 @@ export class Ordering extends AbstractStore<"ordering", TypeOrdering> {
   constructor(state: State) {
     super(state, "ordering");
     this.setServerOrder = this.setServerOrder.bind(this);
+    this.createFolder = this.createFolder.bind(this);
+    this.editFolder = this.editFolder.bind(this);
+    this.deleteFolder = this.deleteFolder.bind(this);
+    this.toggleFolder = this.toggleFolder.bind(this);
+    this.addToFolder = this.addToFolder.bind(this);
+    this.removeFromFolder = this.removeFromFolder.bind(this);
   }
 
   /**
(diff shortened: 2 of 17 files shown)


packages/client/components/state/stores/Ordering.ts
```diff
@@ -89,11 +207,255 @@ export class Ordering extends AbstractStore<"ordering", TypeOrdering> {
...
+  /**
+   * Create a new folder containing the given servers
+   * @param name Folder name
+   * @param serverIds Servers to place in the folder
+   * @returns The new folder's id
+   */
+  createFolder(name: string, serverIds: string[] = []) {
+    const id = `${FOLDER_PREFIX}${Date.now().toString(36)}${Math.random()
+      .toString(36)
+      .slice(2, 8)}`;
```
Dadadah: We have access to ULID generation in the client via the `ulid` package.

```typescript
import { ulid } from "ulid";

const id = `${FOLDER_PREFIX}${ulid()}`;
```

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -212,96 +228,33 @@ export const ServerList = (props: Props) => {
...
-                          <VoiceStatus.Graphic
-                            status={entry.item.voiceStatus}
-                          />
-                        </Show>
-                      </>
-                    }
-                    fallback={entry.item.name}
-                    interactive
+            <Switch>
+              <Match when={entry.item.type === "server" && entry.item}>
```
Dadadah: Shouldn't `entry.item` be before the equality check here? Otherwise if it's falsy the first check might error.

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -212,96 +228,33 @@ export const ServerList = (props: Props) => {
...
+                    server={item().server}
+                    selectedServer={props.selectedServer}
+                    menuGenerator={props.menuGenerator}
                   />
-                </a>
-              </div>
-            </Tooltip>
+                )}
+              </Match>
+              <Match when={entry.item.type === "folder" && entry.item}>
```
Dadadah: Shouldn't `entry.item` be before the equality check here? Otherwise if it's falsy the first check might error.

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -341,6 +294,175 @@ export const ServerList = (props: Props) => {
...
+              size={42}
+              holepunch={collapsed() && mentions() ? "top-right" : "none"}
+              overlay={
+                <Show when={collapsed() && mentions()}>
+                  <Unreads.Graphic count={mentions()} unread />
+                </Show>
+              }
+              fallback={<MdFolder />}
+            />
+          </a>
```
Dadadah: Rather than falling back to the MdFolder icon, could it display the icons of the servers inside? This may require making a new container to hold avatars in a smaller state, but I think it will make it more clear what's in each folder if it shows them.

Dadadah (changes requested):
Excellent first draft. Needs a few changes, but overall I'm happy with how this looks code-wise. Do note that in order to merge this we have to make sure the android client supports it (or atleast doesn't stomp the folders).

This change will likely need buy-in from the core team, so expect merging this to take a little bit.

I believe in it's current state it would not pass UI review, and that is primarily due to the folders not showing their contents when collapsed. I left a comment regarding this.

Dadadah:
To make dnd work you will need to:

* Wrap every server and folder in a container
  * The container wrapping servers should turn into a folder container on consider (this may require updating `Draggable.tsx` to accept an onConsider function)
  * The onChange handler should create a folder for servers, and add the server to the folder for folders
* Wrap each of these containers in a Draggable with type=`servers`
* Disable morphing on all the dragables on the servers list (this will require updating `Draggable.tsx`)

It's possible I may be missing something with this, but this should be it. This may be a tall request, so let me know if you run into any issues and I can take a look at it and help implement it.

GalaxyAkiii:
Hey @Dadadah, thanks for the feedback.

Couldn't sleep last night (or really tonight so far) so ended up staying up overnight hacking away at this even before you left your comment and all of yesterday before practice. A good chunk of what you brought up was actually stuff I had started looking into right after putting up the initial PR, so I'm glad to say it's pretty much all sorted now. I do apologise, by my own standards this was a half done job that shouldn't have been put out in the first place.

Regarding the Android client, that's a really great point about not stomping over folders, thank you for bringing it up. Having tinkered with and tried developing an alternative iOS client in the past (sitting in the ether for now), I know the client codebases fairly well and went back to check specifically from your point today. If folders were stored directly inside `ordering.servers`, any client without folder support (like the Android app) would just see a flat list of server IDs and overwrite the whole array the first time someone reorders a server on their phone, losing all the folder data. So I spent some time today moving folder storage into its own `"server-folders"` sync key instead. That way `ordering.servers` stays as plain string IDs, and older/other clients can read and write to it without accidentally touching any folder settings.

On the drag-and-drop side, I actually started out trying to adapt `Draggable.tsx` originally with `onConsider` and disabling morphing like you had suggested much earlier, but `solid-dnd-directive` kept shifting the DOM elements around mid-drag. Because items kept hopping out of the way as you approached them, trying to drop *onto* a server to create or join a folder felt super finicky. 

I ended up writing a dedicated drag handler for the server rail (`railDrag.ts`) instead. The rail items stay put while dragging and an insertion line shows where you're dropping. If you drop onto the middle of a server it folds them into a folder, dropping above or below reorders, dragging off to the side cancels, and it includes full keyboard navigation (<kbd>Space</kbd> to pick up/drop, arrows to move) plus localized screen reader announcements.

I also added the mini-grid preview on collapsed folders so you can see avatars of what's inside instead of just a generic folder icon. You'll also notice `ServerRoleEditor.tsx` in the diff: rather than copy-pasting the ~200 lines of palette and picker code that was private to the role editor, I lifted it into a shared `ColourPicker` component in `@revolt/ui` for both to use. As a bonus, it also fixes an old bug where clicking 'Custom colour' on roles in Chrome literally did nothing and never showed up (Chrome blocks `.click()` on hidden color inputs, so the shared component overlays the input over the button so native clicks work).

Video of Folders (File was too big for GitHub)
https://youtu.be/silABip5PPo?si=gUiecyHi4Ii9I7Q3

Folder with Colour Picker
Folder with Colour Picker

Mix of collapsed and coloured folders
Mix of collapsed + coloured + etc

Folders with notifications
Screenshot 2026-09-09 at 1 24 58 AM

That should be all from me. Definitely understand the need for patience in this. Please let me know if you need anything further from me. I have tried to go over everything, but if I have missed anything at all please do let me know.

Dadadah:
Please drop the i18n folder from this pr.

This PR grew in scope a bit so I'll sit down and review it some time over the next few days. Thanks for the write up!

GalaxyAkiii:
All done, Thank you.

Dadadah:
I've spoken with the core team about this, and it's come to our attention that we are unable to modify the existing schema of synced store keys. This means that for this pr, it will need to include a migration from the `ordering.servers` string array to a new value, `ordering.serverSidebar` will work. It should read from `ordering.servers` if `ordering.serverSidebar` is undefined, but write to `ordering.serverSidebar` only. This will allow the original server list to work across old clients.

packages/client/components/ui/components/design/ColourPicker.tsx
```diff
@@ -0,0 +1,258 @@
...
+    height: "18px",
+    flexShrink: 0,
+    borderRadius: "50%",
+    marginRight: "var(--gap-sm)",
+  },
+});
+
+const NoColourIndicator = styled(ColourIndicator, {
+  base: {
+    border: "2px solid var(--md-sys-color-outline)",
```
infi: This needs to be dashed, not solid

Image

infi (dismissed):

GalaxyAkiii:
Fair enough, I will continue working on this tonight or tomorrow on the flight. Thank you both.

GalaxyAkiii:
@Dadadah checking this still applies, since the storage changed on the 9th.

Folders no longer live inside `ordering.servers`. They moved to a separate synced key (`server-folders`), and `ordering.servers` stays a plain `string[]`, so its schema is unchanged from main.

What the PR does still do is write to `ordering.servers` when the order changes. If that's the part you want gone, I'll move writes to `ordering.serverSidebar` and fall back to `ordering.servers` when it's undefined.

Two ways I can read that:

1. `serverSidebar` holds the whole folder-aware structure, folders inline  and `server-folders` goes away.
2. `serverSidebar` holds the ordering with folder ids in it, and folder metadata stays in its own key.

Which did you have in mind? And is a separate synced key okay at all, or should all of this sit under `ordering`?

(@infi's dashed border I'll pick up in the same pass.)

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -341,6 +443,359 @@ export const ServerList = (props: Props) => {
...
+          [railHeld]:
+            props.drag.dragging() === props.server.id ||
+            props.drag.carrying() === props.server.id,
+        }}
+        use:floating={props.menuGenerator(props.server)}
+      >
+        <Show when={isInsertionBefore(props.drag, props.server.id)}>
+          <div class={railInsertion} />
+        </Show>
+        <a href={state.layout.getLastActiveServerPath(props.server.id)}>
```
amycatgirl: This anchor needs to have the `draggable="false"` attribute to prevent it from taking precedence over the draggable container.

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -341,6 +443,359 @@ export const ServerList = (props: Props) => {
...
+                holepunch={collapsed() && mentions() ? "top-right" : "none"}
+                overlay={
+                  <Show when={collapsed() && mentions()}>
+                    <Unreads.Graphic count={mentions()} unread />
+                  </Show>
+                }
+                fallback={
+                  <Show
+                    when={collapsed() && props.entry.servers.length}
+                    fallback={<MdFolder />}
```
amycatgirl: Nit: Wouldn't it make more sense to use FolderOpen here?

Preferably, this would use the much clearer `folder_open` icon from material symbols, but this could be done in a separate PR.

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -341,6 +443,359 @@ export const ServerList = (props: Props) => {
...
+            role="button"
+            tabindex="0"
+            aria-expanded={!collapsed()}
+            onClick={() => state.ordering.toggleFolder(props.entry.folder.id)}
+          >
+            <FolderIcon
+              style={{
+                // a box shadow rather than an outline, so that the unread
+                // badge still draws on top of it
+                "box-shadow":
```
amycatgirl: Nit: I should probably leave this for UX Review, but I think that an outline looks out of place with the rest of the UI.

Maybe a background using the colour plus some transparency on top of it could be better? Like when the folder is expanded.

amycatgirl (changes requested):
I was testing this locally and ran into an issue when dragging servers around with the custom draggable implementation. Left a comment about it.

GalaxyAkiii:
Went with option 2 for the migration, keeping `ordering.servers` untouched and handling the sidebar list under a new field on `ordering`.

 Summary of where it stands:

* `ordering.servers` is now read-only fallback. Nothing writes to it anymore, so its schema is untouched and older clients keep the list they already had.
* `ordering.serverSidebar` is a flat `string[]` holding server IDs and folder IDs. A reorder is the only operation that writes to it. Reads fall back to `ordering.servers` while it is undefined, which carries an existing server order across the first time this client touches the list.
* `server-folders` still holds the folder metadata (name, colour, collapsed state, members).

A folder's members are kept in `serverSidebar` directly after the folder's own ID. This preserves placement when unpacking a folder or dragging a server out of one, keeping everything in place rather than dropping it to the bottom of the rail.

`orderedEntries()` resolves a folder at its ID in `serverSidebar`, falling back to the position of its first member otherwise. That fallback also covers an older client overwriting the `ordering` key: folder data in `server-folders` survives, and relative rail position is recovered from where the members sit.

That leaves the question of whether a second synced key is acceptable. If the core team would rather everything sat under `ordering`, folding `server-folders` in is a contained change and I'm happy to do it. Just let me know.

Also in this pass:
* @infi: Restored the dashed border (`2px dashed var(--md-sys-color-on-surface-variant)`) on the default swatch.
* @amycatgirl: Added `draggable="false"` on the server anchor. Links are draggable by default, so the native drag could take the pointer before the rail saw it.
* @amycatgirl: Swapped the expanded folder glyph to `folder_open`. Left the material symbols change for a separate PR as you suggested.
* @amycatgirl: Dropped the ring on collapsed folders. The colour now uses the same translucent background the expanded state already had, so the two match rather than using different treatments for the same thing. Happy to put it back if the ring reads better to you or whoever, since you flagged it as a design call.

Screenshot 2026-09-14 at 8 48 51 PM

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -341,6 +443,341 @@ export const ServerList = (props: Props) => {
...
+    );
+
+  const holdsSelected = () =>
+    props.entry.servers.some((server) => server.id === props.selectedServer());
+
+  return (
+    <FolderGroup
+      expanded={!collapsed()}
+      style={{
+        background: props.entry.folder.colour
```
amycatgirl, line 584: I meant as a background for the avatar container but close enough. 

I meant something more akin to this: 

Image

amycatgirl (approved):
Seems to work completely fine now.

Left a comment about the server folder background while not expanded since there seemed to be a misunderstanding between what I meant and the code I highlighted in the review.

But again, UX review is not my thing, you can feel free to ignore my comment since it's better to leave this to Jennifer.

LGTM

packages/client/src/interface/navigation/servers/ServerList.tsx
GalaxyAkiii: Fair enough, I will take that into consideration and do it as well although I will await UX review and go with which one is selected. Thank you.

packages/client/components/app/menus/ServerContextMenu.tsx
```diff
@@ -233,6 +236,47 @@ export function ServerContextMenu(props: { server: Server }) {
         </ContextMenuSubMenu>
       </Show>
 
+      <ContextMenuSubMenu
+        onClick={() =>
+          state.ordering.createFolder("New Folder", [props.server.id])
+        }
```
Dadadah: By having this be the default behavior you have made touch users unable to add a server to an existing folder. This will likely need to be solved by either:
* Detecting touch support and canceling on touch
  * May require updating the Device hook to add touch detection
* Detecting if the device is on a small enough screen to justify touch only mode
  * Will require using layout from the Device hook.

packages/client/src/interface/navigation/servers/railDrag.ts
```diff
@@ -0,0 +1,428 @@
...
+  /**
+   * Eat the click which follows the pointer release that ended a drag
+   * @param e Click event
+   */
+  function swallowClick(e: MouseEvent) {
+    e.preventDefault();
+    e.stopPropagation();
+    window.removeEventListener("click", swallowClick, { capture: true });
+  }
+
```
Dadadah: This seems to be eating the first click after a successful drag.

https://github.com/user-attachments/assets/060fb3c2-91e3-4ad3-bd0c-cd1e7fa2af20

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -341,6 +443,341 @@ export const ServerList = (props: Props) => {
 /**
  * Server list container
  */
+/**
+ * A single server in the list
+ */
```
Dadadah: Looks like this got duplicated or there is an orphan, it should get cleaned up.

Dadadah (changes requested):
Found a few bugs.

When you drag a server that is earlier in the server list to a server that is later in the server list, the new folder appears at the position of the earlier server, which seems counter intuitive.

https://github.com/user-attachments/assets/18d25623-281b-4250-8f02-00bf4c6dc3ba

The changes to the tabbing seem to have made it where servers are unable to be tab selected - I am only able to tab to server folders.

I am also unable to open a folder by pressing enter - perhaps that's due to missing an href on the link?

Dropping a folder onto its open self causes an unnecessary sync.

Dadadah:
There is a bug in the syncing process that must get fixed before this is merged so I'm gonna get a PR up for that.

GalaxyAkiii:
Apologies for the delay. I am working on another big project, but I should be submitting the final code for this tonight.

GalaxyAkiii:
Pushed an update addressing the remaining review points:
- Prevented touch clicks from triggering folder creation on the submenu item so the menu opens properly on mobile/touch.
- Fixed click getting swallowed on the first click after dropping a drag.
- Fixed folder placement when dragging an earlier server onto a later one.
- Added tabindex and Enter key handler to toggle folders and navigate via keyboard.
- Avoided unnecessary sync when dropping an entry into its current position.
- Updated collapsed folder preview background tint.
- Cleaned up duplicate docblock.

Edit: Updated this to Amelia's UI review although I have all three designs for whichever is selected in review.

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -117,6 +133,133 @@ export const ServerList = (props: Props) => {
...
+                  }))),
+            ]
+          : [{ id: entry.id, kind: "server" as const }],
+      ),
+    container: () => rail,
+    label: (id) => {
+      for (const entry of props.orderedEntries) {
+        if (entry.id === id)
+          return entry.type === "folder"
+            ? entry.folder.name || "Folder"
```
Dadadah: This fallback string should probably be translated.

packages/client/src/interface/navigation/servers/ServerList.tsx
```diff
@@ -339,8 +461,352 @@ export const ServerList = (props: Props) => {
...
+        background:
+          !collapsed() && props.entry.folder.colour
+            ? folderTint(props.entry.folder.colour, 18)
+            : undefined,
+      }}
+    >
+      <Tooltip
+        placement="right"
+        content={props.entry.folder.name || "Folder"}
+        aria={props.entry.folder.name || "Folder"}
```
Dadadah: These should also be translated

Dadadah (changes requested):
This is nearly ready, so please avoid making any large changes so we can get this merged soon. I have a few more review notes.

Alt tabbing while in the keyboard drag mode and clicking back into the window causes you to lose your selection (which is expected) but the server remains in a dragging state.

Video:

https://github.com/user-attachments/assets/5217e239-16bf-476c-8f16-d149a3f8b906

GalaxyAkiii:
Please disregard latest push. Will review when I can. A bit more tired and in a rush sorry.

GalaxyAkiii:
Synced with main and removed the duplicate useLingui declaration in ServerList.tsx. CI is passing now.

Current Design of Folders:
Screenshot 2026-09-20 at 8 30 43 PM

Dadadah:
UI/UX feedback:

1. The folder icon should not have a background.
2. The folder icon should be filled.

Once these are addressed we can merge this pr.

GalaxyAkiii:
Image for reference:
Screenshot 2026-09-21 at 8 43 03 AM


Thanks @Dadadah

Dadadah (approved):

GalaxyAkiii:
My apologies, thank you.

PR #1599, opened 2026-09-08
fix: Make scrolling between channels more consistent

### LINKED ISSUE #1485: scrolling position works between channels
how to reproduce:

1. scroll public logs and voice 2 channels to latest messages
2. scroll one of them up
3. see the other one is also scrolled up

those two channels are chosen because they are not active, but it happens for any channels

arch linux brave browser 1.93.129

Dadadah:
Fix some bugs with channel scrolling in the messages component.

Make the findScrollContainer recursive and search for `auto` in addition to the preexisting `scroll`.
When storing the scroll position to cache, store the scroll position of the parent scrollable element, not the div (which has no scroll). Restore on the parent as well.

Fixes #1485

No UI changes.

## How was this PR tested?

Scrolling up channels and switching channels, before change the scroll position would maintain between channels. After change, the channel always scrolls all the way down, unless you have scrolled high enough to trigger a message load. Note that this was the original intended behavior based on what was written for this component before.

## Checklist:

None

- `main` 
  - \#1599 :point\_left:

### DIFF
--- a/packages/client/components/app/interface/channels/text/Messages.tsx
+++ b/packages/client/components/app/interface/channels/text/Messages.tsx
@@ -189,6 +189,21 @@ export function Messages(props: Props) {
   }
 
   /**
+   * Helper function to find the closest parent scroll container
+   * @param el Element
+   * @returns Element
+   */
+  function findScrollContainer(el: Element | null | undefined) {
+    if (!el) {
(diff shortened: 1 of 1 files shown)


Arsabutispik (approved):
:shipit:

PR #1600, opened 2026-09-08
feat: image cropping

### LINKED ISSUE #139: feat: allow users to crop images / icons in client.
When uploading an image, crop and allow user to confirm.

For the purpose for this issue, just implement a way to crop arbitrary images in app.

Arsabutispik:
I am cropping them images.

Upon selecting a file on: Avatar/Icon, Banner, Emojis a Dialog appears and it gives you an UI to crop the image with the ratio it will be used in the client
Fixes #139

## How was this PR tested?

- [x] I cropped them images

image

</details>

## Checklist:

I cropped the AI away

- `main` 
  - \#1600 :point\_left:

### DIFF
--- a/packages/client/components/app/interface/settings/channel/Overview.tsx
+++ b/packages/client/components/app/interface/settings/channel/Overview.tsx
@@ -18,6 +18,7 @@ import {
   Text,
 } from "@revolt/ui";
 
+import { CropFileInput } from "@revolt/ui/components/design/CropFileInput";
 import { ChannelSettingsProps } from "../ChannelSettings";
 
 /**
@@ -108,10 +109,16 @@ export default function ChannelOverview(props: ChannelSettingsProps) {
           <Text class="label">
             <Trans>Channel Info</Trans>
           </Text>
-          <Form2.FileInput
+          <CropFileInput
             control={editGroup.controls.icon}
             accept="image/*"
+            label={t`Channel Icon`}
+            imageJustify={false}
             maxSize={instance.limits().file_upload_size_limits["icons"]}
+            ratio={1}
+            ratioLabel={t`Square`}
+            allowModeToggle={false}
+            dialogTitle={<Trans>Crop Channel Icon</Trans>}
           />
           <Form2.TextField
             minlength={1}
--- a/packages/client/components/app/interface/settings/server/Overview.tsx
+++ b/packages/client/components/app/interface/settings/server/Overview.tsx
@@ -15,6 +15,7 @@ import {
   Text,
 } from "@revolt/ui";
 
+import { CropFileInput } from "@revolt/ui/components/design/CropFileInput";
 import { ServerSettingsProps } from "../ServerSettings";
 
 /**
@@ -219,14 +220,18 @@ export default function ServerOverview(props: ServerSettingsProps) {
     <Column gap="xl">
       <form onSubmit={submit}>
         <Column>
-          <Form2.FileInput
+          <CropFileInput
             control={editGroup.controls.icon}
             accept="image/*"
             label={t`Server Icon`}
             imageJustify={false}
             maxSize={instance.limits().file_upload_size_limits["icons"]}
+            ratio={1}
+            ratioLabel={t`Square`}
+            allowModeToggle={false}
+            dialogTitle={<Trans>Crop Server Icon</Trans>}
           />
-          <Form2.FileInput
+          <CropFileInput
             control={editGroup.controls.banner}
             accept="image/*"
             label={t`Server Banner`}
@@ -234,6 +239,10 @@ export default function ServerOverview(props: ServerSettingsProps) {
             imageRounded={false}
             imageJustify={false}
             maxSize={instance.limits().file_upload_size_limits["banners"]}
+            ratio={232 / 100}
+            ratioLabel={t`Banner`}
+            allowModeToggle={false}
+            dialogTitle={<Trans>Crop Server Banner</Trans>}
           />
           <Form2.TextField
             minlength={1}
--- a/packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
+++ b/packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
@@ -18,6 +18,7 @@ import {
   Row,
   Text,
 } from "@revolt/ui";
+import { CropFileInput } from "@revolt/ui/components/design/CropFileInput";
 
 /**
  * Emoji list
@@ -76,13 +77,17 @@ export function EmojiList(props: { server: Server }) {
         <Column>
           <Row align>
             <Column>
-              <Form2.FileInput
+              <CropFileInput
                 control={editGroup.controls.file}
                 accept="image/*"
+                label={t`Emoji`}
                 imageJustify={false}
                 allowRemoval={false}
                 maxSize={instance.limits().file_upload_size_limits["emojis"]}
-                hideErrors={true}
+                ratio={1}
+                ratioLabel={t`Square`}
+                allowModeToggle={false}
+                dialogTitle={<Trans>Crop Emoji</Trans>}
               />
             </Column>
             <Column grow>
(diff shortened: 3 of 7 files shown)


Arsabutispik:
I forgot channel and role icons fml

packages/client/components/ui/components/design/CropFileInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+            <ImageCropper
+              src={p().objectUrl}
+              ratio={props.ratio}
+              ratioLabel={props.ratioLabel}
+              allowModeToggle={props.allowModeToggle ?? true}
+              ref={(h) => (cropHandle = h)}
+            />
+          </Dialog>
+        )}
+      </Show>
```
Dadadah: Image cropper should be brought into the modals package

packages/client/components/ui/components/design/ImageCropper.tsx
```diff
@@ -0,0 +1,519 @@
...
+const Segment = styled("button", {
+  base: {
+    appearance: "none",
+    border: "none",
+    background: "transparent",
+    color: "var(--md-sys-color-on-surface)",
+    font: "inherit",
+    fontSize: "14px",
+    fontWeight: 500,
+    padding: "8px 16px",
```
Dadadah, line 177: Please use the gap params here

Dadadah (changes requested):
I'm not a fan of how this component works and it's role in this ecosystem. I think a better way to implement this would be to have a "mutator" for files that are uploaded to the file picker, and have that mutator able to modify the file in the file picker.

For example, instead of changing the file input to the crop file input, you can add a `onFiles` handler that allows changing the files (async) in the file input before they are displayed. Then inside this `onFiles` handler you can call the dialog you made here and return the cropped image back to the input.

packages/client/components/app/interface/settings/channel/Overview.tsx
```diff
@@ -108,10 +109,16 @@ export default function ChannelOverview(props: ChannelSettingsProps) {
...
+          <CropFileInput
             control={editGroup.controls.icon}
             accept="image/*"
+            label={t`Channel Icon`}
+            imageJustify={false}
             maxSize={instance.limits().file_upload_size_limits["icons"]}
+            ratio={1}
+            ratioLabel={t`Square`}
+            allowModeToggle={false}
+            dialogTitle={<Trans>Crop Channel Icon</Trans>}
```
Dadadah: This change moved the image input from center aligned to left aligned, not sure if that's intentional. It also added a label but that's probably good.

packages/client/components/app/interface/settings/server/emojis/EmojiList.tsx
```diff
@@ -76,13 +77,17 @@ export function EmojiList(props: { server: Server }) {
...
             <Column>
-              <Form2.FileInput
+              <CropFileInput
                 control={editGroup.controls.file}
                 accept="image/*"
+                label={t`Emoji`}
                 imageJustify={false}
                 allowRemoval={false}
                 maxSize={instance.limits().file_upload_size_limits["emojis"]}
-                hideErrors={true}
```
Dadadah, line 85: This was hiding errors for a reason - this particular form should only show errors in one place, under the submit button. Should probably re-add this.

packages/client/components/ui/components/design/ImageCropper.tsx
```diff
@@ -0,0 +1,519 @@
...
+export type CropMode = "ratio" | "freeform";
+
+export interface CropResult {
+  blob: Blob;
+  dataUrl: string;
+  width: number;
+  height: number;
+}
+
+export interface ImageCropperHandle {
```
Dadadah, line 24: This interface is duplicated at line 193

packages/client/components/ui/components/design/ImageCropper.tsx
```diff
@@ -0,0 +1,519 @@
...
+    const sh = r.h * scale;
+
+    const canvas = document.createElement("canvas");
+    canvas.width = Math.round(sw);
+    canvas.height = Math.round(sh);
+    const ctx = canvas.getContext("2d");
+    if (!ctx) throw new Error("ImageCropper: failed to get canvas context");
+    ctx.drawImage(imgRef, sx, sy, sw, sh, 0, 0, canvas.width, canvas.height);
+
+    const type = props.outputType ?? "image/png";
```
Dadadah: Output crop type is always png. This results in files being larger than their original. The file size limit also only occurs on the original, not the cropped file.
## Repro
1. Try to add a channel icon that is a jpeg and send it an image that is near the size limit (2.5mb) 
2. The crop window will come up. Execute the crop.
3. Submit the form
4. Image will now fail to upload due to being too large.

I reproduced this in main vs this branch and confirmed.

The cropped image also keeps its old file name, even though it has been changed to png.

Dadadah (changes requested):
I have some more comments for ya that I found while testing this out.

The x button on the input doesn't work anymore (it doesn't remove the image)

packages/client/components/ui/components/design/ImageCropper.tsx
Arsabutispik: I am unable to find which ones I am supposed to use, may I get some guidance

Arsabutispik:
Other than CSS params, I believe everything mentioned should be fixed

packages/client/components/ui/components/design/ImageCropper.tsx
Dadadah: I think the reason they are there in the first place is to enforce all gaps to be the same size - what that means is that you should pick the closest one and use that one. So you'd use `--gap-lg` for the 16px (it'd be 15px) and `--gap-md` for 8px. You can see the gap definitions in `packages/client/components/ui/themes/stoatWebTheme.ts`

Dadadah: wait a minute they're deprecated? Oh then I guess nevermind...

packages/client/components/ui/components/utils/cropProcessor.ts
Dadadah, line 1: This file should probably be named "CropProcessor` not imageProcessor. It is an image processor but there could potentially be additional processors in the future?

Arsabutispik: We could always create a new folder for imageProcessors if we get more than one, for now it makes sense to name it cropProcessor

Dadadah (approved):

packages/client/components/app/interface/settings/server/Overview.tsx
```diff
@@ -234,6 +240,11 @@ export default function ServerOverview(props: ServerSettingsProps) {
             imageRounded={false}
             imageJustify={false}
             maxSize={instance.limits().file_upload_size_limits["banners"]}
+            process={cropProcess({
+              ratio: 232 / 100,
+              ratioLabel: t`Banner`,
+              dialogTitle: <Trans>Crop Server Banner</Trans>,
```
Arsabutispik, line 246: Oops forgot to make these use *t* the component breaks when we use <Trans>

packages/client/components/app/interface/settings/server/Overview.tsx
```diff
@@ -225,6 +226,11 @@ export default function ServerOverview(props: ServerSettingsProps) {
             label={t`Server Icon`}
             imageJustify={false}
             maxSize={instance.limits().file_upload_size_limits["icons"]}
+            process={cropProcess({
+              ratio: 1,
+              ratioLabel: t`Square`,
+              dialogTitle: <Trans>Crop Server Icon</Trans>,
```
Arsabutispik, line 232: Same here

Arsabutispik (commented):
Merge blocker:

Dadadah:
I've added circular crop clipping.

I don't really like the freeform toggle - it's not to material spec. Is it even needed? Is there a benefit to allowing freeform? I think we can just allow the emojis to have their own ratio as that's the only place that makes sense.

Dadadah:
image

PR #1619, opened 2026-09-16
fix: image/video lightbox not using context menu

amycatgirl:
We should probably split the file part of the MessageContextMenu at some point, because it might get unwieldy.

Fixes stoatchat/for-desktop#273

## How was this PR tested?

- [x] Context menu appears correctly
- [x] User is able to copy the file into the clipboard
- [x] User is able to copy a link to the file/gif
- [x] User is able to save a file locally
- [x] User is able to open a file/gif in a new tab

### Copy option

[Screencast From 2026-09-15 21-46-44.webm](https://github.com/user-attachments/assets/430ba5e6-4279-4aae-bfbc-280c568df429)

### Context Menu on GIFs

image

### fungus

image

</details>

## Checklist:

- [ ] I have confirmed that any new dependencies are strictly necessary

No AI was used to write this pull request.

- `main` 
  - \#1619 :point\_left:


packages/client/components/app/menus/MessageContextMenu.tsx
```diff
@@ -28,6 +28,8 @@ import MdShield from "@material-design-icons/svg/outlined/shield.svg?component-s
 
 import MdSentimentContent from "@material-symbols/svg-400/outlined/sentiment_content.svg?component-solid";
 
+import { t } from "@lingui/core/macro";
```
Dadadah, line 31: I don't think we can import t like this - you MUST use `useLingui` or whatever it's called.

packages/client/components/app/menus/MessageContextMenu.tsx
```diff
@@ -148,14 +151,55 @@ export function MessageContextMenu(props: {
...
+    let url: string = "";
+
+    if (props.file instanceof File) {
+      url = props.file.previewUrl;
+    } else if (
+      props.file instanceof ImageEmbed ||
+      props.file instanceof VideoEmbed
+    ) {
+      url = props.file.url;
+    }
```
Dadadah, line 163: This duplicated code should be in a helper function.

packages/client/components/app/menus/MessageContextMenu.tsx
```diff
@@ -307,8 +361,9 @@ export function MessageContextMenu(props: {
             <Trans>Report message</Trans>
           </ContextMenuButton>
         </Show>
-        <ContextMenuDivider />
         <Show when={state.settings.getValue("advanced:admin_panel")}>
+          <ContextMenuDivider />
+
```
Dadadah, line 366: Moving this has caused a missing divider on the normal (no files) context menu.

Image

packages/client/components/modal/modals/ImageViewer.tsx
```diff
@@ -32,22 +33,23 @@ export function ImageViewerModal(
...
           panzoom = zoom;
 
+          ref.addEventListener("pointerdown", (e) => {
+            if (e.button !== 2) {
+              panzoom.handleDown(e);
+              return;
+            }
+          });
+          ref.addEventListener("pointermove", panzoom.handleMove);
+          ref.addEventListener("pointerup", panzoom.handleUp);
```
Dadadah, line 51: Listeners never cleaned up

packages/client/components/app/menus/MessageContextMenu.tsx
```diff
@@ -166,22 +210,32 @@ export function MessageContextMenu(props: {
...
+            </ContextMenuButton>
+          </a>
+          <ContextMenuButton
+            icon={MdContentCopy}
+            onClick={() => copyFile((props.file as File)!.originalUrl)}
+          >
+            <Trans>Copy File</Trans>
           </ContextMenuButton>
-        </a>
+        </Show>
```
Dadadah, line 234: Is this path ever reached? I don't see a way to save a file. When I upload a pdf it just does a text embed with nothing in it and right clicking it does nothing.

packages/client/components/app/menus/MessageContextMenu.tsx
```diff
@@ -166,22 +210,32 @@ export function MessageContextMenu(props: {
     <ContextMenu>
       <Show when={props.file}>
         <ContextMenuButton icon={MdOpenInNew} onClick={openFile}>
-          <Trans>Open file</Trans>
+          <Trans>Open File</Trans>
```
Dadadah, line 213: We lose translations by doing this - is it necessary?

Dadadah (changes requested):
I found a few bugs, evidence attached.

If you are panning when your cursor is on the edge of the image you will lose panability until your cursor goes back on the image.

https://github.com/user-attachments/assets/858fb77b-2843-4024-8aa4-1c4dc799041c

Can you disable panning while the context menu is up? Kind of a bit confusing.

packages/client/components/app/menus/MessageContextMenu.tsx
amycatgirl: I think that's because we don't register a context menu for those embeds. It only gets reached inside the image lightbox.

amycatgirl: That and I couldn't get GIF download from the proxied URL to work properly.

packages/client/components/app/menus/MessageContextMenu.tsx
amycatgirl: I completely forgot that this has an uppercase transform on the CSS side.

packages/client/components/app/menus/MessageContextMenu.tsx
```diff
@@ -148,14 +151,58 @@ export function MessageContextMenu(props: {
...
+        }),
+      ]);
+
+      snackbar.show({
+        message: t`Copied file to clipboard`,
+      });
+    } catch (error) {
+      showError(error);
+    }
+  }
```
Pecacheu, line 188: The problem with your approach to copying images to the clipboard is that a custom clipboard data type to bypass the image format limitation means that you can now **ONLY** paste the copied image back into Stoat, not into any other application, which will surely heavily confuse users (some of whom prefer to the clipboard as their main way of getting images from one place to another.)

What other IM apps typically do to accomplish this is simply forcibly convert any image to a PNG and then copy it to the clipboard. Afaik the real reason for this bizarre limitation isn't so much browser security as the linked article claims, it's actually a hard OS-level limitation in Windows and possibly other OSes, that the clipboard only supports a very limited set of image formats, primarily PNG these days since JPG isn't supported (and I believe BMP is, but no one uses BMP anymore.) You'll notice that not only webapps but **nearly all** apps on Windows mysteriously convert images to PNG when they copy them to the clipboard regardless of their source format, with the exception of when it's copied as file data rather than image data (eg. how File Explorer does it). Anyhow long story short, all that is to say that conversion is actually standard practice even if it may feel like a strange solution.

Luckily, there's a clever dependency-free way to do it on web using a Canvas to convert:

```suggestion
      const blob = await res.blob();

      if (ClipboardItem.supports(blob.type)) {
        await _writeBlob(blob);
      } else {
        if (blob.type.startsWith("image/")) {
          //Convert to PNG
          const c = document.createElement("canvas"),
            ctx = c.getContext("2d"),
            img = new Image();

          img.onload = () => {
            c.width = img.width;
            c.height = img.height;
            ctx!.drawImage(img, 0, 0);
            URL.revokeObjectURL(img.src);
            c.toBlob(_writeBlob, "image/png");
          };
          img.onerror = showError;
          img.src = URL.createObjectURL(blob);
        } else {
          // Workaround for the clipboard API being extremely picky
          // https://developer.chrome.com/blog/web-custom-formats-for-the-async-clipboard-api
          await _writeBlob(blob, `web ${blob.type}`);
        }
      }
    } catch (error) {
      showError(error);
    }
  }

  async function _writeBlob(blob: Blob | null, type?: string) {
    if (!blob) return;
    await navigator.clipboard.write([
      new ClipboardItem({ [type ?? blob.type]: blob }),
    ]);
    snackbar.show({ message: t`Copied file to clipboard` });
  }
```

Side note: IMO it would also be nice to enable the copy image functionality for images people sent as a link, which do show the context menu but currently don't have the copy image option because it isn't a file. Sure, you can always open up the original link and copy it with the native context menu from there, but it would be a convenient time-saver just to build it in, and should be a small change.

Pecacheu (changes requested):

PR #1638, opened 2026-09-20
fix: channel tags on custom instances 

awalone995:
It's a single line fix .
TL;DR:
image


Dadadah (approved):

Dadadah (changes requested):
Revoking my approval - I don't think this works.

awalone995:
> Revoking my approval - I don't think this works.

Did you test from a selfhost instance or stoat.chat? I suspect it's something with url origin logic that can break if you test from a stoat.chat subdomain specifically but idk

Dadadah:
When you post stoat.chat links into your self host it says "Private Channel" which is wrong.

Dadadah:
I think this code was supposed to be part of #999 but it got pulled into #1315

Dadadah:
Please see #1649. Thanks for your contribution.

PR #1649, opened 2026-09-20
fix(instance): Don't show remote channel on selfhosted for own origin

### LINKED ISSUE #1620: Message Links not working on self host
### What happened?

Message links do not work on self host and just show '# Remote Channel'

My assumption:
[Stoat host is set to "stoat.chat"](https://github.com/stoatchat/for-web/blob/main/packages/client/components/common/lib/env.ts#L1)

[paramsFromPathname does not return correctly](https://github.com/stoatchat/for-web/blob/main/packages/client/components/markdown/plugins/anchors.tsx#L107)

[params.host is therefore set to the stoat host](https://github.com/stoatchat/for-web/blob/main/packages/client/components/markdown/plugins/anchors.tsx#L109)

[is remote statement becomes true](https://github.com/stoatchat/for-web/blob/main/packages/client/components/markdown/plugins/anchors.tsx#L110)

Therefore shows '# Remote Channel' and clicking the link does not work.
Copying the href that this '# Remote Channel' contains and opening it within a new tab does work correctly also navigate me to this message or channel.

Dadadah:
Self hosted servers can't link channels. This fixes that. There was some code merged in the instance PR that should have stayed in #999. This hotfixes that code temporarily.

Fixes #1620

- `main` 
  - \#1649 :point\_left:


packages/client/components/markdown/plugins/anchors.tsx
```diff
@@ -106,22 +105,44 @@ export function RenderAnchor(
...
+                  (channel()?.serverId
+                    ? `/server/${channel()!.serverId}`
+                    : "") +
+                  `/channel/${params.channelId}` +
+                  (params.exactMessage && params.messageId
+                    ? `/${params.messageId}`
+                    : ""),
+                location.origin,
+              ).href;
+        // END HOTFIX
```
amycatgirl: Nit: I _know_ this is a Hotfix but could this please be refactored into a if statement?

In my opinion, this is just awful to look at

packages/client/components/markdown/plugins/anchors.tsx
```diff
@@ -136,6 +157,9 @@ export function RenderAnchor(
               class={internalLink()}
               disabled={props.disabled}
               href={internalUrl()}
+              // HOTFIX: See above
+              target={remote ? "_blank" : void 0}
```
amycatgirl: Kinda confused on why a void callback is used here. Would it break if it used the default value of "_self"? (at least, according to mdn: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#target)

amycatgirl (approved):
LGTM, left a small nit and a small question.

packages/client/components/markdown/plugins/anchors.tsx
Dadadah: nope, I just didn't know `_self` was the default.

Dadadah: Actually it turns out doing `_self` breaks the link and forces a reload, doing `void 0` makes the link behave as expected and navigates to a channel without reloading

